PR #3, opened 2024-12-05
feat: Added electrical heating wheel 

energinet-ajw:
Added electrical heating wheel. 
Added Mandaloiren som CODEOWNER.


x-platformcoder (approved):
lgtm

PR #81, opened 2025-01-09
Add measurements ci workflow with proto compile job

jonasdmoeller:
# Description

## Pull-request quality

- [x] Subsystem tests have been tested (by manually deploying to `dev_002`)


.github/workflows/detect-changes.yml
```diff
@@ -50,6 +50,8 @@ jobs:
               - '.github/workflows/ci-capacity-settlement.yml'
               - '.github/workflows/cd.yml'
               - '.devcontainer/**'
+            measurements:
+              - 'source/measurements/**'
```
ShaBloze, line 54: should protobuf compile always be triggered when there are changes in measurements?

.github/workflows/ci-measurements.yml
```diff
@@ -0,0 +1,23 @@
+name: CI Measurements
```
ShaBloze, line 1: Maybe this should be a CI Protobuf? Not should if it would make sense and how often we should do this

jonasdmoeller: I think it should run as part of the measurements (or what wi call it.. bronze?), but only trigger on changes to the contracts folder.

.github/workflows/detect-changes.yml
jonasdmoeller: Changed to only compile proto on changes to /contracts folder

PR #83, opened 2025-01-10
feat: decimal type added to protobuf contract

x-platformcoder:
# Description

The quantity datatype under Point is defined as double, however, this should be a Decimal(18, 3).
There is also some naming corrections to Unit Enum


jonasdmoeller (approved):

source/measurements/src/contracts/PersistSubmittedTransaction.proto
```diff
@@ -71,12 +72,12 @@ enum MeteringPointType {
...
-  U_MVARH = 3;
-  U_KVARH = 4;
-  U_KW = 5;
-  U_T = 6;
-  U_K3 = 7;
+  U_KW = 2;
+  U_MW = 3;
+  U_MWH = 4;
+  U_TONNE = 5;
+  U_KVARH = 6;
```
prtandrup, line 79: Before I left yesterday, I was under the impression that kvarh was the one ruled out, and not K3?

jonasdmoeller: We decided to go with the business domain namings and not the EDI/market specific names. Makes sense?

PR #96, opened 2025-01-15
feat: Add Core CD

ShaBloze:
# Description

I forgot to add a dispatch deployment for the core workspace.

As I understood, we should put the medallion architecture into the Core workspace. So I am adding the Bronze layer to the Core workspace now. Silver and Gold will be added later on.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -66,7 +66,7 @@ jobs:
     if: ${{ needs.changes.outputs.electrical_heating == 'true' }}
     uses: Energinet-DataHub/.github/.github/workflows/promote-prerelease.yml@v14
     with:
-      release_name_prefix: core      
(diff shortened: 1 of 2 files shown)


.github/workflows/cd.yml
```diff
@@ -60,6 +60,14 @@ jobs:
     with:
       release_name_prefix: capacity_settlement
 
+
+  core_promote_prerelease:
+    needs: changes
+    if: ${{ needs.changes.outputs.electrical_heating == 'true' }}
```
HenrikSommer, line 66:
```suggestion
    if: ${{ needs.changes.outputs.bronze == 'true' }}
```

lanni-energinet (approved):

HenrikSommer (approved):

PR #105, opened 2025-01-16
feat: Add gold packages

ShaBloze:
# Description

- https://github.com/Energinet-DataHub/team-volt/issues/976

## Pull-request quality

- [ ] The title adheres to [this guide](https://github.com/Mech0z/GitHubGuidelines)
- [ ] Tests are written and executed locally


source/gold/src/gold/application/migrations.py
BjarkeMeier, line 1: I would not create application and domain folders for a simple tech package.

source/gold/tests/conftest.py
```diff
@@ -0,0 +1,107 @@
...
+    session = configure_spark_with_delta_pip(
+        SparkSession.builder.config("spark.sql.warehouse.dir", warehouse_location)
+        .config("spark.sql.streaming.schemaInference", True)
+        .config("spark.default.parallelism", 1)
+        .config("spark.rdd.compress", False)
+        .config("spark.shuffle.compress", False)
+        .config("spark.shuffle.spill.compress", False)
+        .config("spark.sql.shuffle.partitions", 1)
+        .config("spark.databricks.delta.allowArbitraryProperties.enabled", True)
+        .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension")
```
BjarkeMeier, line 32: On purpose or just copy-paste?

source/gold/tests/conftest.py
```diff
@@ -0,0 +1,107 @@
...
+
+    _create_schemas(session)
+
+    yield session
+
+    session.stop()
+
+
+@pytest.fixture(scope="session")
+def migrate(spark: SparkSession) -> None:
```
BjarkeMeier, line 61: nit: Thoughts about naming of fixtures: When you inject this fixture it will always have been executed. But the name is imperative, which kinda encourages invocation of the fixture in the test. How about something like "migrated" or "migrations_executed"?

BjarkeMeier: Like the spark fixture which is likewise not named "create_spark".

source/gold/coverage.xml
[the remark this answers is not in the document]
ShaBloze: No that was a mistake

source/gold/src/gold/application/migrations.py
ShaBloze: So this package will include streaming setup in the future :)

source/gold/tests/conftest.py
ShaBloze: was copy-paste.

I have now removed uneccesary options

source/gold/tests/conftest.py
ShaBloze: Good point - I like the `migrations_executed` :-)

source/gold/.python-version
```diff
@@ -0,0 +1 @@
+3.10
```
Gizmooa, line 1: Just a note, but why are we at 3.10? Shouldnt we bump?

source/gold/src/gold/__init__.py
```diff
@@ -0,0 +1,2 @@
+def main() -> None:
```
Gizmooa, line 1: Unnecessary functions in the init

ShaBloze: i like it tho

source/gold/src/gold/domain/constants/database_names.py
```diff
@@ -0,0 +1,3 @@
+﻿class DatabaseNames:
+    # Should probably be an environment variable because it is created in Terraform.
```
Gizmooa, line 2: We should create a tech issue for this, just to do it for all of the layers

source/gold/.python-version
ShaBloze: it looks like it set to 3.10 for alle packages, so I think this is OK for now

source/gold/src/gold/domain/schemas/gold_measurements.py
```diff
@@ -0,0 +1,25 @@
...
+    StructField,
+    StructType,
+    TimestampType,
+)
+
+from gold.domain.constants.column_names.gold_measurements_column_names import (
+    GoldMeasurementsColumnNames,
+)
+
+measurements_gold_schema = StructType(
```
Gizmooa, line 13:
```suggestion
gold_measurements_schema = StructType(
```

Shouldn't we have layer first? We have done this other places, just to align

source/gold/src/gold/domain/constants/database_names.py
ShaBloze: i will create one

source/gold/tests/application/test_migrations.py
```diff
@@ -0,0 +1,14 @@
...
+from pyspark.sql import SparkSession
+
+from gold.domain.constants.database_names import DatabaseNames
+from gold.domain.constants.table_names import TableNames
+from gold.domain.schemas.gold_measurements import (
+    measurements_gold_schema,
+)
+
+
+def test__migrations__should_create_bronze_measurements_table(spark: SparkSession, migrations_executed):
```
Gizmooa, line 11:
```suggestion
def test__migrations__should_create_gold_measurements_table(spark: SparkSession, migrations_executed):
```

source/gold/uv.lock
Gizmooa, line 1: Just wondering, shouldn't this `uv.lock` be a part of `.gitignore`, since it is auto generated?

source/gold/.python-version
Gizmooa: Yea I know, it was also just a note for all packages. For the silver i was also like wtf why does it pick 3.10? 😄

source/gold/uv.lock
BjarkeMeier: No, it should not be ignored. See https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile.

source/gold/src/gold/__init__.py
Gizmooa: What are others doing? I can accept if we are aligning, but Im pretty sure we are usually creating empty ones 😄 But in this PR alone there is some __init__ filled with this, and some with none

Gizmooa (approved):

source/gold/uv.lock
Gizmooa: Cool, new to UV so just wanted to make sure 😄

Gizmooa: But how come it is needed, if it is generated from the `.toml` during `uv sync` or similar commands? I'd assume the CI calls such a command, and generates the `uv.lock`? And then it'd have locked the dependencies? 😄

BjarkeMeier: Merely guessing. You should use the exact same versions as I do. If you generate your lock file a few days after me, then uv may resolve to other versions that didn't exist when I generated my lock file.

Gizmooa: I can see they do write `This file should be checked into version control, allowing for consistent and reproducible installations across machines.`. So we can resolve this 😄

kerbou (approved):

PR #161, opened 2025-01-27
Fix: Aligns package and module names

jpm-halfspace:
# Description

Last week, we altered the names of all python packages in measurements to be prefixed with `opengeh_`. This change meant that there was a difference between the name of the package and the resulting module that the package exposes. For example, the package `opengeh_capacity_settlement` currently exposes a module called `capacity_settlement`. 

This discrepancy cause issues when deploying to Databricks, as the package name specified in **dh3-infrastructure** uses the `opengeh_` prefix, resulting in the following error:

image

To accommodate this error, and ensure consistency between packages and modules. This PR changes the module name of all packages to be consistent with that specified in their respective pyproject.toml.

## Pull-request quality


.github/actions/compile-proto/action.yml
```diff
@@ -1,6 +1,11 @@
 name: Compile Measurements Proto Contracts
 description: Compile protobuf contracts for measurements subsystem
 
+inputs:
+  path:
+    description: Path to the directory containing the proto files
+    required: true
+    default: ./source/bronze/src/bronze/infrastructure/contracts
```
ShaBloze, line 8:
```suggestion
    default: ./source/bronze/src/opengeh_bronze/infrastructure/contracts
```

docs/gold/table_definitions/gold_measurements.md
```diff
@@ -1,6 +1,6 @@
 # Measurements - Gold Table Definition
 
-This table contains measurements which is presented as one measurement value per metering point per row. The table depends on `silver.measurements`.
+This table contains measurements which is presented as one measurement value per metering point per row. The table depends on `opengeh_silver.measurements`.
```
ShaBloze, line 3:
```suggestion
This table contains measurements which is presented as one measurement value per metering point per row. The table depends on `silver.measurements`.
```

I don't think we want to change our database names :-)

jpm-halfspace: Sorry - TBH I basically just ran a substitute 😅

johevemi (dismissed):

ShaBloze (dismissed):
LGTM

.github/actions/compile-proto/action.yml
```diff
@@ -1,16 +1,27 @@
 name: Compile Measurements Proto Contracts
 description: Compile protobuf contracts for measurements subsystem
 
+inputs:
+  path:
+    description: Path to the directory containing the proto files
+    required: true
+    default: ./source/bronze/src/opengeh_bronze/infrastructure/contracts
```
djorgensendk, line 8: why this default value. It seem quite specific to a single use case

jpm-halfspace: It was hardcoded within the action before. I just tried to do a minimal change that would allow the workflow to run. 

I do fully agree though!

ShaBloze (approved):

djorgensendk (approved):

PR #168, opened 2025-01-28
ci: Make CI/CD workflows reusable

jpm-halfspace:
# Description

This PR makes both CI and CD workflows reusable which is done in preparation of moving the workflows to `EnergiNet-DataHub/.github` so that the methodology can be incorporated in other repositories.

Additionally, cd has been refactored so that instead of looking at package changes, we'll simply promote all the releases related to a specific PR. This avoids the need to check for changes in packages as only changed packages creates new prereleases.

## Pull-request quality


.github/actions/upload-release-artifacts/action.yaml
djorgensendk, line 1: as this is specific to python, the action's name should reflect that

.github/workflows/reusable-ci.yml
```diff
@@ -0,0 +1,188 @@
...
+        required: false
+        default: ""
+    secrets:
+      dh3serviceaccount_privatekey:
+        required: true
+        description: Used in ci_base job to authenticate with the DataHub3 service account
+
+env:
+  CORE_PREFIX: core
+  NONCORE_PREFIX: bundle
```
djorgensendk, line 31: How is the reusability of this workflow, if in another repository, e.g. 1 or 3 different prefixes areneeded, and the core-concept doesn't exist in that context?

.github/actions/promote-prerelease/action.yml
```diff
@@ -1,42 +1,38 @@
...
-      id: set_release_name
+    - name: Set release name
       shell: bash
       run: |
-        echo "release_name=${{ inputs.release_name_prefix }}_${{ steps.find_pull_request.outputs.pull_request_number }}" >>$GITHUB_OUTPUT
-    - name: Promote ${{ steps.set_release_name.outputs.release_name }} and update ${{ inputs.release_name_prefix }}_latest
+        PREFIX=$(echo "${{ inputs.release_name }}" | grep -o '^.*_' | sed 's/_$//')
+        LATEST_RELEASE="${PREFIX}_latest"
+        echo "LATEST_RELEASE=$LATEST_RELEASE" >>$GITHUB_ENV
+    - name: Promote ${{ inputs.release_name }} and update ${{ env.LATEST_RELEASE }}
```
djorgensendk, line 16: why making this change?
The concept of release name prefixes is widely used throughout the DH3 repositories

.github/workflows/reusable-ci.yml
jpm-halfspace: IMO, the whole `core` / combined release concept is a bit flawed as a python packages are supposed to be a self-contained units, which this concepts breaks.

However, I get your point. There is no easy way to allow for customisation like that as actions cannot accept objects as inputs. One could allow a json string input, and thereby provide this flexibility but I think usage would be quite cumbersome in that scenario.

.github/actions/promote-prerelease/action.yml
jpm-halfspace: Instead of relying on an input string, I use `gh release list ...` to find prereleases related to a PR. This means that I get the actual name of the prerelease (PR number included). So when I then loop over the discovered prereleases, it's easier to simply provide the name and cut off the PR number to get the latest release.

djorgensendk (approved):

johevemi (approved):

lanni-energinet (approved):

PR #178, opened 2025-01-30
Fix/deployment

jpm-halfspace:
# Description

Relates to: https://github.com/Energinet-DataHub/dh3-environments/pull/1345

## Pull-request quality

- [ ] The title adheres to [this guide](https://github.com/Mech0z/GitHubGuidelines)
- [ ] Tests are written and executed locally


.github/workflows/reusable-cd.yml
```diff
@@ -97,10 +97,19 @@ jobs:
         with:
           release_name: ${{ matrix.inputs.name }}
 
+      - name: Create Dispatch Event name
+        id: create_dispatch_event_name
+        run: |
+          package_name=$(echo "${{ matrix.inputs.name }}" | grep -o '^.*_' | sed 's/_$//')
+          event_name="${{ inputs.team_name }}-${package_name}-deployment-request-domain"
```
djorgensendk, line 104: I assume inputs.team_name == measurements

measurements is the name of a sub system not a team. Can the variable have a better name?

jpm-halfspace: It is, but the variable is used both to set the `event-type` for dispatched events and the notification when a dispatch fails. 

Would `subsystem_name` be better, and in that case what should `team_email` be? Also `subsystem_email`?

.github/workflows/reusable-cd.yml
```diff
@@ -17,15 +17,15 @@ name: CD
 on:
   workflow_call:
     inputs:
-      team_name:
+      subsytem_name:
```
djorgensendk, line 20:
```suggestion
      subsystem_name:
```

jpm-halfspace: tak!

djorgensendk (approved):

PR #181, opened 2025-01-30
refac: Container test capacity settlement

johevemi:
# Description


source/capacity_settlement/tests/container_tests/conftest.py
```diff
@@ -4,8 +4,7 @@
 from testcommon.container_test import DatabricksApiClient
 
 
-# this fixture is only used in CD and needs to be commented out for local testing
-# local testing uses the fixture in tests/conftest.py
+# For the test to be started from your local machine, you need to set the environment variables in a .env file.
```
BjarkeMeier, line 7: What should be the name of the file? Can you add the file as `foo.env.sample` with instructions and placeholders for the required variables?

johevemi: the name of the file should be .env according to the documentation. I will make a sample file called sample.env

source/capacity_settlement/tests/__init__.py
```diff
@@ -1,3 +1,7 @@
 from pathlib import Path
 
+from dotenv import load_dotenv
+
+load_dotenv()
```
BjarkeMeier, line 5: Can you add a line about the reason for having this function call?

johevemi: done

source/capacity_settlement/tests/sample.env
BjarkeMeier, line 1: Btw, is there a good reason to have both `testsession.local.settings.sample.yml` and `sample.env`? Or should we stick to just one of them?

johevemi: `testsession.local.settings.sample.yml` is settings and `sample.env` is for environment variables

BjarkeMeier: I thought the first was for settings (by means of yaml) and the latter was for settings (by means of environment variables) 😉

JesperGomes (approved):

PR #189, opened 2025-01-31
docs: Added docs for Gold Observation Transformation

Gizmooa:
# Description

Added docs for observation time, and fixed ruff format issue.


source/gold/README.md
```diff
@@ -11,3 +11,8 @@ See the [gold measurements](src/opengeh_gold/domain/schemas/gold_measurements.py
 The transformation from the Silver layer to the Gold layer consists of the following steps:
 
 - **Exploding silver `points` column**: Each point in the `points` array is expanded into its own row, flattening the data.
+- **Transforming `observation_time`**: The observation time is derived based on the resolution of the measurement:
+    - **Monthly resolution (`P1M`)**: If the `start_datetime` is the first day of the month, months are added based on the point’s position value. Otherwise, if the position is the first one, `start_datetime` is kept; otherwise, the month is  truncated, and months are added according to the position value.
+    - **Hourly resolution (`PT1H`)**: The `observation_time` is calculated by adding `(position - 1) * 1 hour` to `start_datetime`.
+    - **15-minute resolution (`PT15M`)**: The `observation_time` is calculated by adding `(position - 1) * 15 minutes` to `start_datetime`.
+    - **Fallback**: If none of these conditions are met, `start_datetime` is used as the `observation_time`.
```
ShaBloze, line 18: just a thought - when the fallback is to set `observation_time` to `start_datetime` we would probably not notice it, but I would think that it is an error in the data and that it should not go to gold if it was the case.

My suggestion would be to either throw an exception or set the it to None or something like that.

ShaBloze (approved):
LGTM!

Just a single comment for thought

PR #190, opened 2025-01-31
ci: powershell does not like

jpm-halfspace:
# Description

### DIFF
--- a/.github/workflows/reusable-cd.yml
+++ b/.github/workflows/reusable-cd.yml
@@ -82,7 +82,7 @@ jobs:
     name: Promote prerelease
     runs-on: ubuntu-24.04
     needs: cd_release_matrix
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable-cd.yml
```diff
@@ -82,7 +82,7 @@ jobs:
     name: Promote prerelease
     runs-on: ubuntu-24.04
     needs: cd_release_matrix
-    if: ${{ needs.cd_release_matrix.outputs.matrix != '' }}
+    if: ${{ needs.cd_release_matrix.outputs.matrix != '[]' }}
```
jpm-halfspace, line 85: Apparently we get a string with `"[]"` not `""` when no release can be found

djorgensendk (approved):

PR #223, opened 2025-02-05
feat: "Migrated" table, with batch script to load data from migrations.

slb-halfspace:
# Description

This PR adds a migration script with the definition of the "measurements_bronze.migrated" table, and the script which migrates the data from the migrations domain into the measurements domain. The table is a one-to-one copy of the migrations_silver.time_series table, which includes all historical data migrated through the migrations domain. 

The script alongside it is meant to run as a batch script, with a special-case to also handle a full load of data. Currently the table we wish to migrate is roughly 1 TB in size, hence the full load code leverages partitioning to split the work into X chunks which are more managable. The normal usecase of the script simply uses the "created" timestamp to determine what data is yet to be loaded, and loads that once a day.

## References
Task at Volt: 
https://app.zenhub.com/workspaces/team-volt-60a61059028fd7000e49956e/issues/gh/energinet-datahub/team-volt/927


docs/bronze/table_definitions/bronze_migrated.md
```diff
@@ -0,0 +1,25 @@
...
+| valid_to_date           | TimestampType                               | false    | Represents the time series end interval, which is in UTC time matching the beginning of a date in the `Europe/Copenhagen` timezone. The moment is exclusive              | -                                                                                                              |
+| values                  | ArrayType(StructType(...))                  | false    | Array that contains position, quality and quantity for all values of a given time series                                                                                 | -                                                                                                              |
+| &nbsp;&nbsp; position   | &nbsp;&nbsp;StructField(IntegerType)        | false    | The position the time series value had in the received time series                                                                                                       | Never 0                                                                                                        |
```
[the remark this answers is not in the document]
slb-halfspace: I'll do the same for the other 2

source/bronze/src/opengeh_bronze/infrastructure/repository.py
```diff
@@ -0,0 +1,30 @@
+from pyspark.sql import DataFrame, SparkSession
+from pyspark.sql import functions as F
+from pyspark.sql import types as T
+
+from opengeh_bronze.domain.constants.database_names import DatabaseNames
+from opengeh_bronze.domain.constants.table_names import TableNames, MigrationsTableNames
+
+class Repository:
```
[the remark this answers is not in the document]
slb-halfspace: Agreed, but I am still not sure where to put repositories in general... 
Also don't like having a folder called "infrastructure/migrations" as we already have a migrations folder besides infrastructure ;) I've called it "migration_data" for now.

source/bronze/src/opengeh_bronze/domain/transformations/migrate_from_migrations_transformations.py
```diff
@@ -0,0 +1,40 @@
...
+    repository: Repository,
+) -> datetime:
+    return (
+        repository.read_migrations_silver_time_series()
+        .agg(max(col(BronzeMigratedColumnNames.created)))
+        .collect()[0][0]
+    )
+
+
+def create_chunks_of_partitions(
```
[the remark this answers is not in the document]
slb-halfspace: Good idea, I was unsure where to put it

source/bronze/src/opengeh_bronze/infrastructure/repository.py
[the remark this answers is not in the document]
ShaBloze: we have not agreed on a structure yet, so I understand that it might be a bit confusing

source/core/src/core/bronze/application/batch_scripts/migrate_from_migrations.py
```diff
@@ -0,0 +1,103 @@
...
+def migrate_time_series_from_migrations_to_measurements(
+    migrated_transactions_repository: Optional[MigratedTransactionsRepository] = None,
+    migrations_silver_time_series_repository: Optional[MigrationsSilverTimeSeriesRepository] = None,
+) -> None:
+    spark = spark_session.initialize_spark()
+    migrated_transactions_repository = (
+        MigratedTransactionsRepository(spark)
+        if migrated_transactions_repository is None
+        else migrated_transactions_repository
+    )
```
ShaBloze, line 28: Not sure I understand why this is neccessary - if it is for test purpose (mocking), this is not needed. :-)

source/core/src/core/bronze/domain/schemas/migrated_transactions.py
```diff
@@ -0,0 +1,72 @@
...
+                )
+            ),
+            False,
+        ),
+        StructField(BronzeMigratedTransactionsColumnNames.created_in_migrations, TimestampType(), False),
+        StructField(BronzeMigratedTransactionsColumnNames.created_in_measurements, TimestampType(), False),
+    ]
+)
+
+migrations_silver_time_series_schema = StructType(
```
ShaBloze, line 43: I assume that we have 2 schemas here because this one is the silver schema from "Migrations" subsystem. I think it would be a good idea to add a comment to not be confused. 

Alternative is to move it to another file, but I cannot come up with a good name.

source/core/src/core/bronze/domain/transformations/migrate_from_migrations_transformations.py
```diff
@@ -0,0 +1,20 @@
...
+
+def map_migrations_to_measurements(migrations_data: DataFrame):
+    return (
+        migrations_data.drop("partitioning_col")
+        .withColumnRenamed(
+            MigrationsSilverTimeSeriesColumnNames.created,
+            BronzeMigratedTransactionsColumnNames.created_in_migrations,
+        )
+        .withColumn(BronzeMigratedTransactionsColumnNames.created_in_measurements, current_timestamp())
+    )
```
ShaBloze, line 20: Should we `.select()` the expected columns? 

If for some reason the silver schema changes, then our stream would probably fail because our table schema does not match

source/core/src/core/bronze/infrastructure/config/table_names.py
```diff
@@ -1,2 +1,7 @@
 class TableNames:
     bronze_submitted_transactions_table = "submitted_transactions"
+    bronze_migrated_transactions_table = "migrated_transactions"
+
+
+class MigrationsTableNames:
+    silver_time_series_table = "time_series"
```
ShaBloze, line 7:
```suggestion
# Table is from the `Migrations` subsystem
class MigrationsTableNames:
    silver_time_series_table = "time_series"
```

I know that `Migrations` is in the name, but I think we should make it clear that it is from another subsystem.

source/core/src/core/migrations/database_names.py
```diff
@@ -1,3 +1,7 @@
 class DatabaseNames:
     # Should probably be an environment variable because it is created in Terraform.
     measurements_internal_database = "measurements_internal"
+
+    # TODO: Is this right placement?
+    bronze_database = "measurements_bronze"
+    silver_migrations_database = "migrations_silver"
```
ShaBloze, line 7: Maybe we should make a `Migrations` layer or something like that. So we can put all configuration, repository etc. that connects to the `Migrations` subsystem into the `Migrations` layer. Then we also know that it is from another subsystem

source/core/src/core/bronze/application/batch_scripts/migrate_from_migrations.py
slb-halfspace: Agreed, I have properly mock'ed this now :)

source/core/src/core/bronze/domain/transformations/migrate_from_migrations_transformations.py
slb-halfspace: Agreed, changed in newest commit

source/core/src/core/migrations/database_names.py
slb-halfspace: For other unittests, I ended up needing the silver_migrations_database in catalog_settings, meaning that this specific piece of code was removed. Hopefully it is OK to have it in catalog_settings... 

Probably need to extend it somewhere in dh3-infrastructure then, as a environment variable?

source/core/src/core/bronze/application/batch_scripts/migrate_from_migrations.py
```diff
@@ -0,0 +1,85 @@
...
+    migrated_transactions_repository: MigratedTransactionsRepository,
+    latest_created_already_migrated: datetime,
+) -> None:
+    today = datetime.now().date()
+
+    migrations_data = migrations_silver_time_series_repository.read_migrations_silver_time_series().filter(
+        (col(MigrationsSilverTimeSeriesColumnNames.created) < lit(today))
+        & (col(MigrationsSilverTimeSeriesColumnNames.created) > lit(latest_created_already_migrated))
+    )
+
```
ShaBloze, line 51: I would probably move this into the repository (the filter) and create a new method named something like `read_latest_created_by` (i don't have a better idea right now) - and then call that method here. 

This is to remove the filter functionality from the `application` layer, it this is more of an `orchestrator` layer - hope it makes sense :)

source/core/src/core/bronze/application/batch_scripts/migrate_from_migrations.py
```diff
@@ -0,0 +1,85 @@
...
+        num_chunks,
+    )
+    today = datetime.now().date()
+
+    for chunk in chunks:
+        migrations_data_chunk = migrations_silver_time_series_repository.read_migrations_silver_time_series().filter(
+            (lit(chunk[0]) <= col(MigrationsSilverTimeSeriesColumnNames.partitioning_col))
+            & (col(MigrationsSilverTimeSeriesColumnNames.partitioning_col) <= lit(chunk[-1]))
+            & (col(MigrationsSilverTimeSeriesColumnNames.created) < lit(today))
+        )
```
ShaBloze, line 79: Same as above comment - move the filter into a new method in the repository and call the new method here :)

source/core/src/core/bronze/infrastructure/migration_data/silver_time_series_repository.py
```diff
@@ -0,0 +1,36 @@
...
+from core.bronze.infrastructure.config.table_names import MigrationsTableNames
+from core.settings.catalog_settings import CatalogSettings
+
+
+class MigrationsSilverTimeSeriesRepository:
+    def __init__(
+        self,
+        spark: SparkSession,
+    ) -> None:
+        self._spark = spark
```
ShaBloze, line 12:
```suggestion
        self.spark = spark
```

source/core/src/core/migrations/migration_scripts/20250217150000_create_bronze_migrated_transactions.sql
```diff
@@ -0,0 +1,29 @@
...
+            position INT,
+            quality STRING,
+            quantity DECIMAL(18, 6)
+        >
+    > NOT NULL,
+    created_in_migrations TIMESTAMP NOT NULL,
+    created_in_measurements TIMESTAMP NOT NULL
+)
+USING DELTA
+{SKIP_IN_TEST} CLUSTER BY metering_point_id, transaction_insert_date, valid_from_date, transaction_id
```
ShaBloze, line 25: Is this neccessary to skip in test? I think it works without skipping

source/core/src/core/migrations/substitutions.py
```diff
@@ -7,8 +9,10 @@
 def substitutions() -> dict[str, str]:
     catalog_settings = CatalogSettings()  # type: ignore
     return {
+        "{SKIP_IN_TEST}": "--" if "unittest" in sys.modules.keys() else "",
```
ShaBloze, line 12: What is `sys.modules.keys()`

source/core/tests/bronze/infrastructure/migrations_data/test_migrated_transaction_repositories.py
```diff
@@ -0,0 +1,54 @@
...
+    df = spark.createDataFrame([(date,) for date in dates], ["date"])
+
+    # Act
+    chunks = repo.create_chunks_of_partitions_for_data_with_a_single_partition_col(df, "date", 3)
+
+    # Assert
+    assert len(chunks) == 3
+    assert max([len(chunk) for chunk in chunks]) == 2
+    assert min([len(chunk) for chunk in chunks]) == 1
+
```
ShaBloze, line 34: maybe an assert that the total amount of chunks is equal to the count of `dates`? To make sure that we have all 5 in our chunks

docs/bronze/table_definitions/bronze_migrated.md
```diff
@@ -0,0 +1,25 @@
...
+| metering_point_id       | StringType                                  | false    | The GSRN number that uniquely identifies the metering point                                                                                                              | Exactly 18 digits                                                                                              |
+| type_of_mp              | StringType                                  | false    | -                                                                                                                                                                        | Valid values can be found in opengeh-migrations. |
+| historical_flag         | BooleanType                                 | false    | **True:** when the time series has been updated by a later one and thus made historical / **False:** when the time series is the latest/current one for a given interval | -                                                                                                              |
```
HenrikSommer, line 15:
```suggestion
| type_of_mp              | StringType                                  | false    | -                                                                                                                                                                        | Valid values can be found in opengeh-migrations |
| historical_flag         | BooleanType                                 | false    | **True:** when the time series has been updated by a later one and thus made historical <br> **False:** when the time series is the latest/current one for a given interval | -                                                                                                              |
```

source/core/pyproject.toml
```diff
@@ -16,6 +16,7 @@ stream_silver_to_gold_measurements = "core.gold.entry_points:stream_silver_to_go
 notify_transactions_persisted = "core.bronze.entry_points:notify_transactions_persisted"
 ingest_submitted_transactions = "core.bronze.entry_points:ingest_submitted_transactions"
 stream_submitted_transactions_to_silver = "core.silver.entry_points:stream_submitted_transactions"
+migrate_time_series_from_migrations = "core.bronze.entry_points:migrate_time_series_from_migrations_to_measurements"
```
HenrikSommer, line 19: We migrate from Silver, so I guess the entrypoint should be in `core.silver.entry_points`

source/core/src/core/bronze/application/batch_scripts/migrate_from_migrations.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+    migrated_transactions_repository.write_measurements_bronze_migrated(migrations_data_transformed)
+
+
+# Leverage the transaction_insert_date partitioning to split our work into chunks due to the large amount of data to migrate.
+def full_load_of_migrations_to_measurements(
+    migrations_silver_time_series_repository: MigrationsSilverTimeSeriesRepository,
+    migrated_transactions_repository: MigratedTransactionsRepository,
+    num_chunks=10,
```
HenrikSommer, line 60: Do we have any idea, whether 10 chunks will work for production?

slb-halfspace: The production load is ~1.1TiB right now, splitting it in 10 parts should make it considerably more managable. Could probably bump it to 20, but the higher the value the more overhead. What is ideal is hard to gauge, but the goal is to just have a value that should make it possible to complete without OOM.

slb-halfspace: And since there are no joins in any way, the pressure shouldn't be too hard on the machines itself. That is why I think 10 is a good starting-value, and if it is too low then it can be adjusted.

source/core/src/core/bronze/domain/constants/column_names/bronze_migrated_transactions_column_names.py
```diff
@@ -0,0 +1,21 @@
...
+    transaction_id = "transaction_id"
+    transaction_insert_date = "transaction_insert_date"
+    unit = "unit"
+    status = "status"
+    read_reason = "read_reason"
+    valid_from_date = "valid_from_date"
+    valid_to_date = "valid_to_date"
+    values = "values"
+    created_in_migrations = "created_in_migrations"
+    created_in_measurements = "created_in_measurements"
```
HenrikSommer, line 15:
```suggestion
    created = "created"
```

slb-halfspace: I can see the point in having this be just "created", but could it lead to confusion? I am OK to change it to just created, I am just unsure if it is best.

source/core/src/core/bronze/infrastructure/migration_data/silver_time_series_repository.py
```diff
@@ -0,0 +1,61 @@
...
+        self.source_table_name = MigrationsTableNames.silver_time_series_table
+
+    def read_migrations_silver_time_series(self) -> DataFrame:
+        return self.spark.read.table(f"{self.source_database}.{self.source_table_name}")
+
+    def read_migrations_silver_time_series_written_since_timestamp(
+        self, latest_created_already_migrated: datetime
+    ) -> DataFrame:
+        return self.read_migrations_silver_time_series().filter(
+            (col(MigrationsSilverTimeSeriesColumnNames.created) < lit(datetime.now().date()))
```
HenrikSommer, line 30: Is it by design that we do not read data created _today_?

source/core/src/core/bronze/infrastructure/migration_data/silver_time_series_repository.py
```diff
@@ -0,0 +1,61 @@
...
+            & (col(MigrationsSilverTimeSeriesColumnNames.created) > lit(latest_created_already_migrated))
+        )
+
+    def read_chunk_of_migrations_silver_time_series_partitions(
+        self, start_partition_date, end_partition_date
+    ) -> DataFrame:
+        return self.read_migrations_silver_time_series().filter(
+            (lit(start_partition_date) <= col(MigrationsSilverTimeSeriesColumnNames.partitioning_col))
+            & (col(MigrationsSilverTimeSeriesColumnNames.partitioning_col) <= lit(end_partition_date))
+            & (col(MigrationsSilverTimeSeriesColumnNames.created) < lit(datetime.now().date()))
```
HenrikSommer, line 40: Same as above 👆

source/core/src/core/bronze/infrastructure/migration_data/silver_time_series_repository.py
```diff
@@ -0,0 +1,61 @@
...
+        remainder = len(partitions) % num_chunks
+        chunks = []
+        start = 0
+
+        for i in range(num_chunks):
+            end = start + chunk_size + (1 if i < remainder else 0)
+            chunks.append(partitions[start:end])
+            start = end
+
+        return [chunk for chunk in chunks if len(chunk) > 0]
```
HenrikSommer, line 61: What happens if `len(chunk) <= 0`?

source/core/src/core/bronze/infrastructure/repositories/migrated_transactions_repository.py
```diff
@@ -0,0 +1,39 @@
...
+            data.write.format("delta")
+            .option("mergeSchema", "false")
+            .mode("append")
+            .saveAsTable(f"{self.bronze_database_name}.{self.migrated_transactions_table_name}")
+        )
+
+    def calculate_latest_created_timestamp_that_has_been_migrated(self) -> Optional[datetime]:
+        return (
+            self.read_measurements_bronze_migrated_transactions()
+            .agg(F.max(F.col(BronzeMigratedTransactionsColumnNames.created_in_migrations)))
```
HenrikSommer, line 37: In a sql command I've seen similar statement run for a long time. Maybe it is no longer a problem, using liquid clustering?

source/core/src/core/migrations/migration_scripts/20250217150000_create_bronze_migrated_transactions.sql
```diff
@@ -0,0 +1,29 @@
+CREATE TABLE {bronze_database}.{bronze_migrated_transactions_table}
+(
+    metering_point_id STRING NOT NULL,
+    type_of_mp STRING NOT NULL,
+    historical_flag STRING NOT NULL,
+    resolution STRING NOT NULL,
+    transaction_id STRING,
```
HenrikSommer, line 7: `NOT NULL`?

source/core/src/core/settings/catalog_settings.py
```diff
@@ -24,5 +24,7 @@ class CatalogSettings(BaseSettings):
     silver_database_name: str
     gold_database_name: str
 
+    migrations_silver_database_name: str
```
HenrikSommer, line 27: Let's add this to the comment in lines 9 - 16

source/core/tests/bronze/helpers/builders/migrated_transactions_builder.py
```diff
@@ -0,0 +1,53 @@
...
+        transaction_id: str = "UnitTestingTransaction001",
+        transaction_insert_date: datetime = datetime(2025, 2, 10, 12, 0, 0),
+        unit: str = "KWH",
+        status: int = 0,
+        read_reason: str = "",
+        valid_from_date: datetime = datetime(2025, 2, 8, 23, 0, 0),
+        valid_to_date: datetime = datetime(2025, 2, 9, 23, 0, 0),
+        values: list[tuple[int, str, float]] = [(i, "D01", Decimal(i)) for i in range(24)],
+        created_in_migrations: datetime = datetime(2025, 2, 10, 13, 0, 0),
+        created_in_measurements: datetime = datetime(2025, 2, 10, 14, 0, 0),
```
HenrikSommer, line 29: Remember to change this to `created` as well

source/core/src/core/bronze/infrastructure/migration_data/silver_time_series_repository.py
slb-halfspace: The design idea was to run daily, and as such it might as well be around midnight rather than otherwise. It would probably be better to abstract away and simply say "from now", so I will change it :)

source/core/src/core/bronze/infrastructure/migration_data/silver_time_series_repository.py
slb-halfspace: Then it is an empty chunk, and that is useless to us. 
This list-comprehension just filters away all empty chunks.

source/core/src/core/migrations/migration_scripts/20250217150000_create_bronze_migrated_transactions.sql
slb-halfspace: It is nullable in migrations, so I figure it should be so here too.

slb-halfspace: image

source/core/src/core/bronze/infrastructure/migration_data/silver_time_series_repository.py
slb-halfspace: The problem I had in mind is this:

Let's say we start a full load, and we have 1000 partitions that we split into 10 chunks. While we process the first 9, the silver stream in migrations is still running, and adds data to the partitions we have already read from the first 9 chunks. In silver they will have a created timestamp less than our largest created in measurements, since our 10th partition most likely also contains new data. 

Hence the limitation of the "today filter", but it has been changed to just be based on the repository which is nicer :)

HenrikSommer (dismissed):

HenrikSommer (dismissed):

HenrikSommer (approved):

PR #232, opened 2025-02-06
feat: Updated types for Calculation Execute

cenh-halfspace:
# Description

Fixed some types that were not changed.


source/electrical_heating/src/opengeh_electrical_heating/domain/calculation.py
```diff
@@ -13,6 +13,13 @@
 from opengeh_electrical_heating.domain.column_names import ColumnNames
 from opengeh_electrical_heating.domain.types import NetSettlementGroup
 from opengeh_electrical_heating.domain.types.metering_point_type import MeteringPointType
+from opengeh_electrical_heating.infrastructure.electricity_market.child_metering_points.wrapper import (
+    ChildMeteringPoints,
+)
+from opengeh_electrical_heating.infrastructure.electricity_market.consumption_metering_point_periods.wrapper import (
+    ConsumptionMeteringPointPeriods,
+)
+from opengeh_electrical_heating.infrastructure.measurements.measurements_gold.wrapper import TimeSeriesPoints
```
BjarkeMeier, line 22: I've added these public symbols to `infrastructure/__init__.py` to avoid having to know the inner structure when used from outside `infrastructure`. It also adds the benefit of greatly simplified imports.
```suggestion
from opengeh_electrical_heating.infrastructure import (
    ChildMeteringPoints,
    ConsumptionMeteringPointPeriods,
    TimeSeriesPoints,
)
```

source/electrical_heating/tests/scenario_tests/conftest.py
```diff
@@ -10,12 +10,19 @@
 from opengeh_electrical_heating.infrastructure.electricity_market.child_metering_points.schema import (
     child_metering_points_v1,
 )
+from opengeh_electrical_heating.infrastructure.electricity_market.child_metering_points.wrapper import (
```
BjarkeMeier, line 13: Simplify imports as per my previous comment.
Note that I didn't add the schemas/contracts to `infrastructure/__init__.py` as I don't consider them public (they are not used outside infrastructure in production code). So they remain "deep path" imports in tests.

PR #234, opened 2025-02-06
refac: Rename "measurements_bronze" to "calculated_measurements"

JesperGomes:
# Description

Calculated measurements will not be stored in bronze - instead it will go into a dedicated table. This PR renames from "measurements_bronze" to "calculated_measurements". This has no real consequence since measurments are not yet written to any table.

The schema will be updated in a separate PR according a not yet agreed contract with EDI and Measurements Core

## References

https://app.zenhub.com/workspaces/mandalorian-5fd22a1a59e4740018f5ab5a/issues/gh/energinet-datahub/team-mandalorian/529


source/electrical_heating/src/opengeh_electrical_heating/infrastructure/measurements/calculated_measurements/schema.py
```diff
@@ -21,7 +21,7 @@
 
 # Electrical heating result is delivered to Measurements
 # as one transaction per metering point id.
-measurements_bronze_v1 = t.StructType(
+calculated_measurements_v1 = t.StructType(
```
BjarkeMeier, line 24: Versioning a table? Is this because you want to defer adding a view until deemed needed?

JesperGomes: I was focused on renaming, so I didn't really think about it. I will remove _v1

BjarkeMeier (approved):

PR #240, opened 2025-02-07
Test : include unchanges calculated capacity settlement measurements

Troels36:
# Description

Test : that the value that has been calculated is included in the new calculation for capacity settlement, even if the value is the same.


source/capacity_settlement/tests/scenario_tests/wip_Include_unchanges_calculated_measurements/test_scenario.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+from testcommon.dataframes import (
```
jpm-halfspace, line 2: This could be bumped to use the combined `geh_common` package. `testcommon` is available under `geh_common.testing`

jpm-halfspace (approved):

source/capacity_settlement/tests/scenario_tests/wip_Include_unchanges_calculated_measurements/test_scenario.py
jpm-halfspace: However, this could also be done in another PR

PR #250, opened 2025-02-10
feat: Add Calculations and Ten largest quantities

johevemi:
# Description


source/capacity_settlement/src/opengeh_capacity_settlement/domain/calculation.py
```diff
@@ -37,12 +39,19 @@ def execute_core_logic(
     spark: SparkSession,
     time_series_points: DataFrame,
     metering_point_periods: DataFrame,
+    orchestration_instance_id: UUID,
     calculation_month: int,
     calculation_year: int,
     time_zone: str,
+    execution_time: str,
 ) -> CalculationOutput:
     calculation_output = CalculationOutput()
```
BjarkeMeier, line 48: I would instantiate at the end (of the function) when all elements are available. That makes it very easy to see how the output is instantiated and whether something is missing. You'll also be able to remove alle the `| None = None`parts of the output class. This again makes the code less error prone.

source/capacity_settlement/src/opengeh_capacity_settlement/domain/calculation.py
```diff
@@ -52,8 +61,26 @@ def execute_core_logic(
...
+    ten_largest_quantities = time_series_points_ten_largest_quantities.select(
+        ColumNames.metering_point_id,
+        ColumNames.quantity,
+        ColumNames.observation_time,
+    )
+    calculation_output.ten_largest_quantities = ten_largest_quantities
+
     time_series_points = _average_ten_largest_quantities_in_selection_periods(
-        time_series_points, metering_point_periods
+        time_series_points_ten_largest_quantities, grouping
```
BjarkeMeier, line 83: `time_series_points` is now something different from what it was at the beginning of the function (line 62). Reusing names this way makes it way harder to understand the code and to refactor it.

BjarkeMeier: I however don't consider it a problem if you reuse the same name multiple times in consequitive lines. Kinda like chaining. But that's different from using the same name different places in the function.

source/capacity_settlement/tests/scenario_tests/given_a_smoke_test_scenario/then/ten_largest_quantities.csv
```diff
@@ -0,0 +1,11 @@
+metering_point_id;quantity;observation_time;
```
BjarkeMeier, line 1: It's probably because I don't know capacity settlement. How do you arrive at these large quantities given the much smaller time series values?

source/capacity_settlement/src/opengeh_capacity_settlement/domain/calculation.py
johevemi: Move to the bottom of the function and remove none

source/capacity_settlement/tests/scenario_tests/given_a_smoke_test_scenario/then/ten_largest_quantities.csv
johevemi: the ten largest quantities are found in _ten_largest_quantities_in_selection_periods
I don't know if that answers your question.

source/capacity_settlement/src/opengeh_capacity_settlement/domain/calculation.py
```diff
@@ -37,12 +39,19 @@ def execute_core_logic(
...
     calculation_year: int,
     time_zone: str,
+    execution_time: str,
 ) -> CalculationOutput:
     calculation_output = CalculationOutput()
 
+    calculation_output.calculations = spark.createDataFrame(
+        [(str(orchestration_instance_id), calculation_year, calculation_month, execution_time)],
+        schema="orchestration_instance_id STRING, year INT, month INT, execution_time STRING",
+    )
```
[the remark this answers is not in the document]
johevemi: I agree. We discussed how calculations should not even be in execute_core_logic since it is not core logic, and we would eventually move it out. For now, I just created a function for creating calculations.

source/capacity_settlement/src/opengeh_capacity_settlement/domain/calculation.py
johevemi: I made the naming more explicit

BjarkeMeier (approved):

PR #258, opened 2025-02-11
chore: use geh_common

jpm-halfspace:
# Description


source/geh_calculated_measurements/src/geh_calculated_measurements/opengeh_electrical_heating/infrastructure/electricity_market/consumption_metering_point_periods/wrapper.py
```diff
@@ -1,5 +1,5 @@
+from geh_common.pyspark.data_frame_wrapper import DataFrameWrapper
```
johevemi, line 1: There are a lot of places where import order and general structure is change, do you have other setting than the default from our devcontainer? Because when i save it will move the imports in the order it was.

johevemi (approved):

PR #273, opened 2025-02-16
chore: Align devcontainer naming

BjarkeMeier:
# Description

Align renaming, by replacing this:
image

with this:
image


energinet-ajw:
When was it decided to call it "calculated"?

BjarkeMeier:
> When was it decided to call it "calculated"?

I don't know whether this is about decisions. I see it as aligning names. And I find "core" and "calculations" work well together (otherwise I think it should be "core_measurements" and "calculated_measurements" - but I don't see any value in those longer names).

energinet-ajw (approved):

PR #276, opened 2025-02-16
refac(calculated): Do not use `.withColumn()`

BjarkeMeier:
# Description


source/geh_calculated_measurements/src/geh_calculated_measurements/opengeh_capacity_settlement/domain/calculation.py
```diff
@@ -82,7 +83,7 @@ def execute_core_logic(
     ten_largest_quantities = time_series_points_ten_largest_quantities.select(
         ColumNames.metering_point_id,
         ColumNames.quantity,
-        ColumNames.observation_time,
+        CalculatedNames.observation_time_hourly,
```
johevemi, line 86: why

johevemi (dismissed):

source/geh_calculated_measurements/src/geh_calculated_measurements/opengeh_capacity_settlement/domain/calculation.py
```diff
@@ -204,26 +210,22 @@ def _explode_to_daily(
...
+    return (
+        df.select(
+            "*",
+            F.explode(
+                F.sequence(
+                    F.lit(calculation_start_date.date()), F.lit(calculation_end_date.date()), F.expr("interval 1 day")
+                )
+            ).alias("date_local"),
+        )
+        .select("*", F.to_utc_timestamp("date_local", time_zone).alias(ColumNames.date))
```
JesperGomes, line 222: do we know if this makes any practical difference in performance?

BjarkeMeier: No, but it's recommended by "the internet" and JP :)

BjarkeMeier: https://blog.devgenius.io/why-to-avoid-multiple-chaining-of-withcolumn-function-in-spark-job-35ee8e09daaa

BjarkeMeier: https://bytesofdeepankar.hashnode.dev/stop-withcolumn-chain

BjarkeMeier:
Tech refinement decided that we don't worry about using `.withColumn()` as long as we only have few and not chained.

PR #281, opened 2025-02-18
style: Align flake8 line-length with ruff setting

HenrikSommer:
# Description

Part of (2) _Standardized Projects_ in [Maturing python development](https://energinet.atlassian.net/wiki/spaces/D3/pages/1279328262/Maturing+python+development)

To avoid flake8 creating squigglies for lines > 79 chars in VS Code the setting is aligned with the current ruff setting of 120 chars


.editorconfig
BjarkeMeier, line 1: Why do we have/use both flake8 and ruff?

HenrikSommer: Not sure, maybe @jpm-halfspace knows?

jpm-halfspace: We don't need to use both flake8 and ruff. `ruff` implements all the rules from flake8

BjarkeMeier: Then I'd prefer that we remove flake8 rather than configure it to align with ruff.

.vscode/settings.json
```diff
@@ -3,5 +3,8 @@
         "."
     ],
     "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true
+    "python.testing.pytestEnabled": true,
+    "flake8.args": [
```
jpm-halfspace, line 7: This is not required. If you have the flake8 extension installed, you can safely uninstall this and install the ruff extension instead.

.editorconfig
HenrikSommer: flake8 is still in use in _opengeh-migration_ but I guess we can do without the extension

HenrikSommer:
We remove/disable the flake8 VS Code extension

PR #284, opened 2025-02-19
refac: Move execution time to where it is used

johevemi:
# Description


source/geh_calculated_measurements/src/geh_calculated_measurements/capacity_settlement/domain/calculation.py
```diff
@@ -96,8 +94,8 @@ def _create_calculations(
     orchestration_instance_id: UUID,
     calculation_month: int,
     calculation_year: int,
-    execution_time: str,
 ) -> DataFrame:
+    execution_time = datetime.now(UTC).strftime("%Y-%m-%dT%H:%M:%SZ")
```
JesperGomes, line 98: why is the type string in line 108? Should it be timestamp?

johevemi: doesn't it need to be a string because of the formating we are using? if i don't have strftime on it will give a timestamp like this 2025-02-19 11:43:13.87856

BjarkeMeier: And that's the proper time ;-) You can truncate if using datetime functionality if you don't want the ticks.

source/geh_calculated_measurements/src/geh_calculated_measurements/capacity_settlement/domain/calculation.py
BjarkeMeier, line 1: This PR invalidates the work of @energinet-ajw that enabled testing execution time as well in scenario tests.

source/geh_calculated_measurements/src/geh_calculated_measurements/capacity_settlement/domain/calculation.py
johevemi: don't we need T and Z like this YYYY-MM-DDTHH:MM:SSZ

johevemi: see [adr-144](https://energinet.atlassian.net/wiki/spaces/D3/pages/1295483020/ADR+144+-+CSV+file+format+for+use+within+DataHub)

johevemi: this will be the format 2025-02-19 11:43:13

PR #291, opened 2025-02-20
feat: [Core] Move receipt flow to silver layer

HenrikSommer:
# Description

When we receive data in the bronze table, we are not yet sure that we can unpack the protobuf data and that it contains valid data, so we change our receipt flow to read from the silver table.

## References

[Volt 1055](https://app.zenhub.com/workspaces/team-volt-60a61059028fd7000e49956e/issues/gh/energinet-datahub/team-volt/1055)


source/core/src/core/bronze/domain/transformations/submitted_transactions_transformation.py
```diff
@@ -7,7 +7,7 @@
     ValueColumnNames,
 )
 from core.bronze.domain.constants.descriptor_file_names import DescriptorFileNames
-from core.bronze.infrastructure.helpers.path_helper import get_protobuf_descriptor_path
+from core.utility.path_helper import get_protobuf_descriptor_path
```
Gizmooa, line 10: Just a consideration, is it a bit messy having protobuf path getter inside a path_helper? Since it is only used by 2-ish workflows? We could easily end up having a lot of different path_helpers, could we generalize it somehow in a nice way?

Gizmooa: Could we maybe have a some sort of Protobuf helper? Do we have enough things to put in here, such that it makes sense?

source/core/src/core/bronze/infrastructure/streams/kafka_stream.py
```diff
@@ -1,12 +1,10 @@
...
-from core.bronze.infrastructure.settings import (
-    KafkaAuthenticationSettings,
-    StorageAccountSettings,
-    SubmittedTransactionsStreamSettings,
-)
+from core.bronze.infrastructure.settings import SubmittedTransactionsStreamSettings
 from core.settings.catalog_settings import CatalogSettings
+from core.settings.kafka_authentication_settings import KafkaAuthenticationSettings
+from core.settings.storage_account_settings import StorageAccountSettings
 from core.utility.shared_helpers import get_checkpoint_path
```
Gizmooa, line 10: Just a general thought, not to anything you've changed. 

```
checkpoint_location = get_checkpoint_path(
            self.data_lake_settings, self.bronze_container_name, "submitted_transactions"
        )
```

on line 21, it would make more sense to simply set this inside `__init__` such that we dont have to initialize `checkpoint_location` for every method call. This way we dont need to set `data_lake_setting` and `bronze_container_name` as class parameters. But instead have `submitted_transactions_checkpoint_location` or similar

Gizmooa: ^This is general, this could also be done the other places where we initialize checkpoint_location inside the method

source/core/src/core/silver/infrastructure/streams/process_manager_stream.py
```diff
@@ -0,0 +1,28 @@
...
+        self.data_lake_settings = StorageAccountSettings().DATALAKE_STORAGE_ACCOUNT  # type: ignore
+        self.silver_container_name = CatalogSettings().silver_container_name  # type: ignore
+
+    def write_stream(
+        self,
+        dataframe: DataFrame,
+    ):
+        checkpoint_location = get_checkpoint_path(
+            self.data_lake_settings, self.silver_container_name, "processed_transactions"
+        )
```
Gizmooa, line 23: Same as other checkpoint comment

source/core/src/core/silver/infrastructure/streams/process_manager_stream.py
```diff
@@ -0,0 +1,28 @@
...
+        event_hub_instance = KafkaAuthenticationSettings().event_hub_instance  # type: ignore
```
Gizmooa, line 24: Is the # type:ignore needed here?

source/core/src/core/silver/infrastructure/streams/submitted_transactions_repository.py
```diff
@@ -0,0 +1,21 @@
...
+        self.silver_database_name = CatalogSettings().silver_database_name  # type: ignore
+
+    def read_submitted_transactions(self) -> DataFrame:
+        return (
+            self.spark.readStream.format("delta")
+            .option("ignoreDeletes", "true")
+            .table(f"{self.silver_database_name}.{SilverTableNames.silver_measurements}")
+            .filter(
+                f"{SilverMeasurementsColNames.orchestration_type} = 'OT_SUBMITTED_MEASURE_DATA'"
+            )  # TODO: Add this to constants
```
Gizmooa, line 20: Missing todo, and I agree ;-)

source/core/tests/bronze/infrastructure/settings/test_submitted_transactions_stream_settings.py
```diff
@@ -8,7 +8,7 @@ def test__submitted_transactions_stream_settings__continuous_streaming_should_be
     os.environ["CONTINUOUS_STREAMING_ENABLED"] = "true"
 
     # Act
-    actual = SubmittedTransactionsStreamSettings()
+    actual = SubmittedTransactionsStreamSettings()  # type: ignore
```
Gizmooa, line 11: Is this needed?

Gizmooa: (the ignore)

source/core/src/core/contracts/process_manager/persist_submitted_transaction_proto_version.py
```diff
@@ -0,0 +1,2 @@
+class PersistSubmittedTransactionProtoVersion:
+    version = "v1"
```
Gizmooa, line 2: What are we going to do here, when v2? Change this value? Shouldnt this maybe be called `version_1` such that we can distinguish and roll forward/backwards?

source/core/src/core/silver/infrastructure/streams/process_manager_stream.py
HenrikSommer: Let's do this in another PR

Gizmooa (approved):

PR #298, opened 2025-02-21
fix: typo in devcontainer

jpm-halfspace:
# Description

### DIFF
--- a/.devcontainer/calculated/devcontainer.json
+++ b/.devcontainer/calculated/devcontainer.json
@@ -9,6 +9,7 @@
   "workspaceFolder": "/workspace/source/geh_calculated_measurements",
   "postStartCommand": "uv sync",
   "containerEnv": {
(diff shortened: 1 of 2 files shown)


.devcontainer/calculated/devcontainer.json
```diff
@@ -57,7 +58,7 @@
           "tests"
         ],
         "python.analysis.include": [
-          "/workspace/source/ /**/*"
+          "/workspace/source/**/*"
```
BjarkeMeier, line 61: Is this also a problem in other devcontainers? (including other repos)

BjarkeMeier (approved):

.devcontainer/calculated/devcontainer.json
jpm-halfspace: I'll check

PR #321, opened 2025-02-27
Exploratory script for generating simulated electricity data

JakobStricker:
# Description

Include a folder named 'Exploratory' that includes a script that generates simulated data for Child_metering_points, Consumption_metering_point_periods, and Time_series_points. The data is generated based on assumption on how actual data is distributed from the production environment.


source/geh_calculated_measurements/src/geh_calculated_measurements/exploratory/electrical_heating/electrical_heating_simulation.py
BjarkeMeier, line 1: How is this intended to be executed?
It doesn't look like production code and thus shouldn't be included in the package.

JakobStricker:
No need for the code to be in production

PR #323, opened 2025-02-27
refac: Moved wait for condition function to tests/common

energinet-ajw:
Tech. debt from another PR. Two fixtures used the same wait function.
The function is moved to tests/common.


source/geh_calculated_measurements/tests/common/utils/log_query_client_wrapper.py
```diff
@@ -0,0 +1,38 @@
+import time
+from datetime import timedelta
+
+from azure.identity import DefaultAzureCredential
+from azure.monitor.query import LogsQueryClient, LogsQueryResult, LogsQueryStatus
+
+
+class LogQueryClientWrapper:
```
BjarkeMeier, line 8: Move to `geh_common.testing`?

BjarkeMeier: Someone told me to not refactor but do it right the first time 😜

BjarkeMeier (approved):

PR #361, opened 2025-03-05
refac: Calculation measurements migration

johevemi:
# Description

image

## References

https://github.com/Energinet-DataHub/dh3-environments/actions/runs/13675032562/job/38233651617

### DIFF
--- a/source/geh_calculated_measurements/src/geh_calculated_measurements/database_migrations/settings/catalog_settings.py
+++ b/source/geh_calculated_measurements/src/geh_calculated_measurements/database_migrations/settings/catalog_settings.py
@@ -1,7 +1,8 @@
-from pydantic_settings import BaseSettings
+from geh_common.application.settings import ApplicationSettings
+from pydantic import Field
 
 
-class CatalogSettings(BaseSettings):
+class CatalogSettings(ApplicationSettings):
     """Configuration class inheriting pydantic's BaseSettings to automatically load environmental variable.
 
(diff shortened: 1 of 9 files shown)


source/geh_calculated_measurements/src/geh_calculated_measurements/database_migrations/settings/catalog_settings.py
```diff
@@ -10,7 +11,7 @@ class CatalogSettings(BaseSettings):
     catalog_name (str): The name of the unity catalog created in infrastructure.
     """
 
-    catalog_name: str
+    catalog_name: str = Field(init=False)
 
     class Config:
         case_sensitive = False
```
BjarkeMeier, line 17: Remove?

johevemi: Why?

BjarkeMeier: Why keep it? 😉 
It should not be needed when using ApplicationSettings.

johevemi: I can see that 'case_sensitive = False' is not part of Application Settings and I do not know if 'case_sensitive = False' is needed or not

PR #364, opened 2025-03-05
feat: Append to Gold when not exists 

ShaBloze:
# Description

## References

- https://github.com/Energinet-DataHub/team-volt/issues/1009


source/core/src/core/silver/infrastructure/repositories/silver_measurements_repository.py
```diff
@@ -19,6 +20,10 @@ def __init__(self) -> None:
         self.data_lake_settings = StorageAccountSettings().DATALAKE_STORAGE_ACCOUNT
         self.silver_container_name = SilverSettings().silver_container_name
 
+    def read(self) -> DataFrame:
```
HenrikSommer, line 23:
```suggestion
    def read_stream(self) -> DataFrame:
```

HenrikSommer (approved):

PR #367, opened 2025-03-06
fix: devcontainers in WSL

jpm-halfspace:
# Description


.devcontainer/Dockerfile
BjarkeMeier, line 1: When testing this locally I get the following error while trying to open in devcontainer:

image

jpm-halfspace: Container is already in use (on your system) - guessing you already have it open on windows :)

johevemi (approved):

PR #371, opened 2025-03-06
feat: Add submitted transaction enum validations

HenrikSommer:
# Description

Add validation for all enumerations in `PersistSubmittedTransaction.proto` and document.

## References

[volt 1064](https://app.zenhub.com/workspaces/team-volt-60a61059028fd7000e49956e/issues/gh/energinet-datahub/team-volt/1064)


source/core/src/core/silver/domain/validations/enum_validations.py
```diff
@@ -5,29 +5,48 @@
...
+    ]
+    return F.col(SubmittedTransactionsUnpackedColumnNames.orchestration_type).isin(valid_orchestration_types)
 
 
 def validate_quality_enum() -> Column:
     """Quality check: QCST01-02."""
-    return F.lit(True)
+    valid_qualities = [name for name in Quality._member_names_ if name != Quality.Q_UNSPECIFIED.value]
+
+    return is_not_null(SubmittedTransactionsUnpackedColumnNames.points).otherwise(
```
ShaBloze, line 27: this `is_not_null` got me confused. I would think that if `is_not_null` is true, then we should check the quality of all points, but it seems that it returns when the points is not null? maybe it is just me?

HenrikSommer: If the `points` column is null the function returns `false`, meaning data in the column is invalid.
Otherwise, the `quality` attribute is validated for all points in the array contained in the `points` column.

I agree, it is not easily readable, can we come up with a better name?

HenrikSommer: I have renamed the function to `column_is_not_null()` which is slightly more readable, and created a test: `test__quality_enum_validations_when_points_column_is_null__then_data_is_invalid()`

source/core/src/core/silver/domain/validations/enum_validations.py
```diff
@@ -5,29 +5,48 @@
...
     """Quality check: QCST01-02."""
-    return F.lit(True)
+    valid_qualities = [name for name in Quality._member_names_ if name != Quality.Q_UNSPECIFIED.value]
+
+    return column_is_not_null(SubmittedTransactionsUnpackedColumnNames.points).otherwise(
+        F.forall(
+            F.col(SubmittedTransactionsUnpackedColumnNames.Points.quality),
+            lambda x: x.isNotNull() & x.isin(valid_qualities),
+        )
+    )
```
ShaBloze, line 32: When look at the `column_is_not_null` method, it looks like we are checking if the column `isNull` and not `isNotNull`. 
```suggestion
    return F.when(is_null(SubmittedTransactionsUnpackedColumnNames.points), F.lit(False)).otherwise(
        F.forall(
            F.col(SubmittedTransactionsUnpackedColumnNames.Points.quality),
            lambda x: x.isNotNull() & x.isin(valid_qualities),
        )
    )
```

I think what confuses me is the `.otherwise` after the `column_is_not_null`.

To me, it looks like if the column is not null, then we return (?). Otherwise (if null), we check if all the points quality is valid.

ShaBloze (approved):

PR #391, opened 2025-03-11
refac: Silver and Gold stream

ShaBloze:
# Description

## References

- https://github.com/Energinet-DataHub/team-volt/issues/1076


source/core/src/core/gold/infrastructure/config/spark.py
```diff
@@ -3,9 +3,5 @@
 
 
 def initialize_spark() -> SparkSession:
-    spark_conf = (
-        SparkConf(loadDefaults=True)
-        .set("spark.sql.session.timeZone", "UTC")
-        .set("spark.sql.shuffle.partitions", "auto")
-    )
+    spark_conf = SparkConf(loadDefaults=True).set("spark.sql.session.timeZone", "UTC")
```
Gizmooa, line 6: I think `.set("spark.sql.shuffle.partitions", "auto")` will be necessary, or some custom value - since the default is 200 which is extremely low.

source/core/src/core/gold/infrastructure/repositories/gold_measurements_repository.py
```diff
@@ -2,55 +2,59 @@
...
-            .foreachBatch(batch_operation)
         )
 
-        if terminate_on_empty:
-            df_write_stream.trigger(availableNow=True).start().awaitTermination()
-        else:
-            df_write_stream.start().awaitTermination()
+        stream_settings = StreamingSettings()
+
+        if stream_settings.continuous_streaming_enabled is False:
```
Gizmooa, line 40:
```suggestion
        if StreamingSettings().continuous_streaming_enabled is False:
```
I think this would be cleaner, since we're not using the `stream_settings` in other places beneath

source/core/src/core/utility/delta_table_helper.py
```diff
@@ -0,0 +1,27 @@
+from delta import DeltaTable
+from pyspark.sql import DataFrame, SparkSession
+
+
+def append_if_not_exists(spark: SparkSession, dataframe: DataFrame, table: str, merge_columns: list[str]) -> None:
```
Gizmooa, line 5: Not the biggest fan of append as a part of the naming, as it is a merge behind the functionality. merge_if_not_exists, or something similar?

source/core/tests/gold/application/streams/test_gold_measurements_stream.py
```diff
@@ -1,42 +1,73 @@
...
 
 
 def test__stream_measurements_silver_to_gold__calls_expected(spark: SparkSession):
     # Arrange
     silver_repo_mock = Mock(spec=SilverMeasurementsRepository)
-    sut.SilverMeasurementsRepository = Mock(return_value=silver_repo_mock)
     gold_repo_mock = Mock(spec=GoldMeasurementsRepository)
-    sut.GoldMeasurementsRepository = Mock(return_value=gold_repo_mock)
-    silver_repo_mock.read_stream.return_value = Mock()
+    with (
```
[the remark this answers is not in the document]
ShaBloze: No we have not :P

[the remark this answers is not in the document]
Gizmooa: Oooof, because we have already started to diverge xD

source/core/src/core/gold/infrastructure/repositories/gold_measurements_repository.py
```diff
@@ -2,55 +2,57 @@
...
     def append_if_not_exists(self, gold_measurements: DataFrame) -> None:
         """Append to the table unless there are duplicates based on all columns except 'created'.
 
         :param gold_measurements: DataFrame containing the data to be appended.
         """
-        existing_data = gold_measurements.sparkSession.table(self.table)
-        new_data = gold_measurements.alias("new_data")
-
-        condition = [new_data[column] == existing_data[column] for column in self._merge_columns()]
+        spark = spark_session.initialize_spark()
```
Gizmooa, line 48: Isn't it overhead to initialize a spark session for every call to `append_if_not_exists`?

ShaBloze: A new spark session should not be created for every call, as the `initialize_spark` calls a `getOrCreate()`. I am not sure if that will result in an overhead?

Gizmooa: It is still a function call, but I see what you mean - but anyhow, the method code could be simplified by initializing it in the `__init__`

ShaBloze: ah yes, I will move it to the __init__ :D

Gizmooa (approved):

Gizmooa (approved):

PR #392, opened 2025-03-11
feat: Write to calculations and ten_largest_quantities

johevemi:
# Description


source/geh_calculated_measurements/src/geh_calculated_measurements/common/infrastructure/repository.py
```diff
@@ -15,22 +16,48 @@ def __init__(
         catalog_name: str | None = None,
     ) -> None:
         self._spark = spark
-        self._calculated_measurements_database_name = CalculatedMeasurementsInternalDatabaseDefinition.DATABASE_NAME
-        self._calculated_measurements_table_name = CalculatedMeasurementsInternalDatabaseDefinition.MEASUREMENTS_NAME
         self._catalog_name = catalog_name
+
```
JesperGomes, line 20: This file (repository.py) is in common because it contains code that is shared between the different calculations, but "calculations" and "ten_largest_quantities" are only relevant for capacity settlement. Wouldn't it be better to keep capacity-settlement-specific code under capacity_settlement?

johevemi: I placed it there because it uses the same database/schema, but it could be moved to capacity settlement to follow the business logic. I will look into moving it and still keep it generic

JesperGomes: ok - you could consider if your new repository should then propagate to common if the type 'CalculatedMeasurements'

JesperGomes (approved):

PR #394, opened 2025-03-11
Add measurements client

jonasdmoeller:
# Description


.github/workflows/detect-changes.yml
```diff
@@ -9,13 +9,16 @@ on:
     outputs:
       dotnet:
         value: ${{ jobs.changes.outputs.dotnet }}
+      packages:
```
ShaBloze, line 12:
```suggestion
      nuget-packages:
```

maybe? to not confuse with python packages

docs/Measurements.Client/Documentation.md
ShaBloze, line 1: should there be an example of how to use it?

lanni-energinet (dismissed):

source/dotnet/Measurements.Client/MeasurementsClient.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        var response = await _httpClient
+            .GetAsync(
+                $"/measurements?MeteringPointId={query.MeteringPointId}&StartDate={query.Date}&EndDate={query.Date.Plus(Duration.FromDays(1))}",
+                cancellationToken)
+            .ConfigureAwait(false);
+
+        if (!response.IsSuccessStatusCode)
+        {
+            return null;
```
ShaBloze, line 28: are we sure that we just want to return `null` here?

jonasdmoeller: Done

docs/Measurements.Client/Documentation.md
jonasdmoeller: Not for the time being. As the packages are only used internally on datahub projects I think we should avoid too much documentation. I basically copy pasted process manager 🙂 
[](https://github.com/Energinet-DataHub/opengeh-process-manager/blob/main/docs/ProcessManager.Client/Documentation.md)

ShaBloze (approved):

lanni-energinet (approved):

PR #403, opened 2025-03-12
Core/Feat: Transform values to silver and align with GehCore

Gizmooa:
# Description


source/core/src/core/contracts/process_manager/enums/resolution.py
```diff
@@ -5,3 +5,4 @@ class Resolution(Enum):
     R_UNSPECIFIED = "R_UNSPECIFIED"
     R_PT15M = "R_PT15M"
     R_PT1H = "R_PT1H"
+    R_P1M = "R_P1M"
```
Gizmooa, line 8: Not sure if this should be included(?) Can we receive 1m from PM?

ShaBloze (approved):

PR #409, opened 2025-03-12
Expand unit and quality

jonasdmoeller:
# Description

### DIFF
--- a/source/dotnet/Measurements.Application/Responses/GetMeasurementResponse.cs
+++ b/source/dotnet/Measurements.Application/Responses/GetMeasurementResponse.cs
@@ -58,6 +58,7 @@ public class GetMeasurementResponse
             "measured" => Quality.Measured,
             "estimated" => Quality.Estimated,
             "calculated" => Quality.Calculated,
+            "missing" => Quality.Missing,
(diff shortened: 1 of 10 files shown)


source/dotnet/Measurements.Application/Responses/GetMeasurementResponse.cs
```diff
@@ -66,8 +67,13 @@ private static Unit ParseUnit(string unit)
...
         {
-            "KWH" => Unit.KWh,
-            "MWH" => Unit.MWh,
+            "kwh" => Unit.Kwh,
+            "kw" => Unit.Kw,
+            "mw" => Unit.Mw,
+            "mwh" => Unit.Mwh,
+            "tonne" => Unit.Tonne,
+            "kvarh" => Unit.Kvarh,
+            "mvar" => Unit.Mvar,
```
ShaBloze, line 76: should the strings her match the unit defined her? https://github.com/Energinet-DataHub/opengeh-python-packages/blob/main/source/geh_common/src/geh_common/domain/types/quantity_unit.py

jonasdmoeller: Fixed

PR #420, opened 2025-03-13
refac: Use `@testing` decorator from `geh_common`

BjarkeMeier:
# Description

See [here](https://github.com/Energinet-DataHub/opengeh-measurements/pull/420#pullrequestreview-2681861140).


BjarkeMeier:
It will be replaced by another upcoming PR.

PR #439, opened 2025-03-17
test: Enable parallel execution of tests to improve speed

jpm-halfspace:
# Description

## Sequential Tests #440 
image

## 2 Workers (-n auto --dist load)
image


source/geh_calculated_measurements/tests/conftest.py
```diff
@@ -57,25 +61,81 @@ def clear_cache(spark: SparkSession) -> Generator[None, None, None]:
...
+def fix_print():
+    """
+    pytest-xdist disables stdout capturing by default, which means that print() statements
+    are not captured and displayed in the terminal.
+    That's because xdist cannot support -s for technical reasons wrt the process execution mechanism
+    https://github.com/pytest-dev/pytest-xdist/issues/354
+    """
+    original_print = print
+    with mock.patch("builtins.print") as mock_print:
+        mock_print.side_effect = lambda *args, **kwargs: original_print(*args, **{"file": sys.stderr, **kwargs})
```
BjarkeMeier, line 137: So this "redirects" the stdout from `print(...)` to stderr?

source/geh_calculated_measurements/tests/conftest.py
```diff
@@ -57,25 +61,81 @@ def clear_cache(spark: SparkSession) -> Generator[None, None, None]:
     spark.catalog.clearCache()
 
 
+def _get_spark():
```
BjarkeMeier, line 64: This is massive. Are there reasons for all the settings? Do we want the same setup in all repos?
If so - should we move this to geh_common?

source/geh_calculated_measurements/tests/electrical_heating/electrical_heating_tests/infrastructure/measurements/conftest.py
```diff
@@ -28,12 +29,42 @@ def measurements_gold_with_data(spark: SparkSession) -> None:
...
-            ("123456789012345", "consumption", datetime.datetime(2022, 1, 1, 0, 0, 0), Decimal("10.500")),
-            ("123456789012345", "consumption", datetime.datetime(2022, 1, 1, 1, 0, 0), Decimal("12.750")),
-            ("223456789012345", "electrical_heating", datetime.datetime(2022, 1, 1, 0, 0, 0), Decimal("25.000")),
-            ("323456789012345", "supply_to_grid", datetime.datetime(2022, 1, 1, 0, 0, 0), Decimal("5.250")),
-            ("423456789012345", "consumption_from_grid", datetime.datetime(2022, 1, 2, 0, 0, 0), Decimal("8.125")),
-            ("523456789012345", "net_consumption", datetime.datetime(2022, 1, 2, 0, 0, 0), Decimal("15.000")),
+            (
+                "123456789012345",
+                "consumption",
+                datetime.datetime(2022, 1, 1, 0, 0, 0, tzinfo=timezone.utc),
```
BjarkeMeier, line 35: Just curious: Why does the tzinfo matter to this PR?

source/geh_calculated_measurements/tests/conftest.py
jpm-halfspace: Yes - pytest-xdist doesn't support the `-s` flag that allow you to use `print()` in your tests. However, pytest always prints `stderr` so this is a work around.

source/geh_calculated_measurements/tests/electrical_heating/electrical_heating_tests/infrastructure/measurements/conftest.py
jpm-halfspace: Because I set the default timezone for spark to be UTC. So `tzinfo` is required because when dates are loaded into spark, they'll otherwise be transformed from the local time zone (which differs depending on the machine that tests are running on) into UTC

source/geh_calculated_measurements/tests/conftest.py
jpm-halfspace: We could move this to `geh_common`, but I wanted to do it as two separate steps. Also, it depends what other people think...

BjarkeMeier: Perhaps we can discuss it at the tech refinement.

source/geh_calculated_measurements/tests/electrical_heating/electrical_heating_tests/infrastructure/measurements/conftest.py
BjarkeMeier: Makes sense 👍

source/geh_calculated_measurements/tests/conftest.py
jpm-halfspace: pytest-xdist disables stdout capturing by default, which means that print() statements
    are not captured and displayed in the terminal.
    That's because xdist cannot support -s for technical reasons wrt the process execution mechanism
    https://github.com/pytest-dev/pytest-xdist/issues/354

Doesn't this already explain it?

source/geh_calculated_measurements/tests/conftest.py
```diff
@@ -35,25 +38,34 @@ def clear_cache(spark: SparkSession) -> Generator[None, None, None]:
...
-        # Enable Hive support for persistence across test sessions
-        .config("spark.sql.catalogImplementation", "hive")
-        .enableHiveSupport()
-    )
-    session = configure_spark_with_delta_pip(session).getOrCreate()
-    yield session
-    session.stop()
+    yield _spark
+    _spark.stop()
+    shutil.rmtree(data_dir)
```
BjarkeMeier, line 54: Perhaps we don't want to delete to save time when running consecutive test sessions locally.
But considering the problems on my other branch, let's try to see if this would solve that problem.

BjarkeMeier (approved):

PR #453, opened 2025-03-18
feat: Remove is_deleted, and keep is_cancelled in gold"

slb-halfspace:
# Description

This PR adds the "is_cancelled" column to gold measurements. 
Said column is then used in the capacity and electrical heating views to not have any cancelled data.
At the same time, it removes the existing "is_deleted" column from silver, as it is redundant.
Finally, it also improves "StreamingSettings" to allow for setting maxBytesPerTrigger and maxFilesPerTrigger

## References
https://app.zenhub.com/workspaces/team-volt-60a61059028fd7000e49956e/issues/gh/energinet-datahub/team-volt/1102


source/core/src/core/gold/domain/constants/enums/statuses.py
ShaBloze, line 1: We are not using "status" - are we? Should this be deleted?

source/core/src/core/migrations/migration_scripts/20250318160000_add_is_cancelled_to_gold_measurements.sql
ShaBloze, line 1: Should we set "is_deleted" for all and not just `orchestration_type != "migration"`

to avoid having Nulls - and maybe we should have a constraint that it cannot be null.

I don't know if we can identify rows from silver in Bronze, but if we can we can set the `is_cancelled` correct

source/core/src/core/migrations/migration_scripts/20250319130000_update_capacity_settlement_view_with_where.sql
```diff
@@ -0,0 +1,20 @@
...
+        *,
+        ROW_NUMBER() OVER (PARTITION BY metering_point_id, observation_time ORDER BY transaction_creation_datetime DESC) AS row_num
+    FROM {gold_database}.{gold_measurements}
+    WHERE metering_point_id IS NOT NULL
+    AND quantity IS NOT NULL
+    AND observation_time IS NOT NULL
+    AND metering_point_type IS NOT NULL
+    AND metering_point_type IN ('consumption', 'capacity_settlement')
+    AND is_cancelled IS NOT NULL 
+    AND NOT is_cancelled
```
ShaBloze, line 16: if we add constraint that it cannot be null, this `is not null` is not neccessary

source/core/src/core/settings/streaming_settings.py
```diff
@@ -12,6 +15,17 @@ class StreamingSettings(BaseSettings):
...
+    maxBytesPerTrigger: Optional[str] = Field(default=None, init=False)
+
+    def apply_streaming_settings(self, streaming_writer: DataStreamWriter) -> DataStreamWriter:
+        if self.continuous_streaming_enabled is False:
+            streaming_writer = streaming_writer.trigger(availableNow=True)
+        if self.maxFilesPerTrigger is not None:
+            streaming_writer = streaming_writer.option("maxFilesPerTrigger", self.maxFilesPerTrigger)
+        if self.maxBytesPerTrigger is not None:
+            streaming_writer = streaming_writer.option("maxBytesPerTrigger", self.maxBytesPerTrigger)
+        return streaming_writer
```
ShaBloze, line 28: I don't think that this should be a environment variable for the StreamingSettings, as this is generel settings for streaming, but instead environment variable specific for a Stream. So if we want to set something for the Gold Stream, we should have a `GoldStreamingSetting` and use that in the Gold Stream. Does it make sense?

I guess we want to have it as an environment variable because it is easier to change in all the environments and it might differ in the environments?

source/core/src/core/gold/domain/constants/enums/statuses.py
slb-halfspace: I made it before we decided on a final strategy, it wasn't used, so yeah it is now deleted 😅

source/core/src/core/migrations/migration_scripts/20250318160000_add_is_cancelled_to_gold_measurements.sql
slb-halfspace: I kept it as only non-migration as we don't know whether data in gold from migrations is cancelled or not, and I'd rather not risk marking a cancelled thing as not-cancelled. But realistically speaking, this shouldn't be an issue so I've changed it to your suggestion :)

slb-halfspace: Except the part about identifying rows in silver from bronze, I don't think that is a good way of doing it. I'd rather delete all migration data and re-run the bronze -> silver migration stream from scratch.

slb-halfspace: But that is only a problem in T001, where we have already taken care of it for like 99.999% of the data so it shouldn't be an issue at all. Don't know about Prod, if we even have something running there?

source/core/src/core/settings/streaming_settings.py
slb-halfspace: My thinking is that previously I implemented something in that style, aka MigratedStreamingSettings, but was asked to remove it in favour of the StreamingSetting. Plus, the two will be fighting over the same environment_variable names, so I figured this is OK to have generally. 

It is something we realistically want for every single stream, so I don't see why we wouldn't want it generally.

ShaBloze: But don't we want to have different `maxBytesPerTrigger` and `maxFilesPerTrigger` depending on the stream?

slb-halfspace: Are you suggesting we name the environment variables "GOLD_STREAM_MAX_BYTES_PER_TRIGGER" to differentiate? I was under the impression that we would most likely have one cluster per stream anyway, and then it can be pretty clean to just have it as a general one.

ShaBloze (approved):

PR #474, opened 2025-03-21
Update geh_common dependency to version 5.4.5

BjarkeMeier:
Upgrade the geh_common package to the latest version to ensure compatibility and access to recent features and fixes.


energinet-ajw:
Does the lambda expression select the data frame to be written?
testing(selector = lambda x:  x.something)

energinet-ajw (approved):

BjarkeMeier:
> Does the lambda expression select the data frame to be written? testing(selector = lambda x: x.something)

Exactly 👍

PR #477, opened 2025-03-21
test: Activated subsystem step 4 for electrical heating and capacity settlement

energinet-ajw:
Activated subsystem step 4 for electrical heating and capacity settlement.
Minor refac to base_test_fixture.


BjarkeMeier (approved):
Did you test this on dev_002/003?

PR #487, opened 2025-03-24
feat: add current view v1

x-platformcoder:
# Description

Access to the latest (non-cancelled) measurements for all metering points
So that is possible to calculated gaps/missing data periods that must be sent to the Market Actor

## References

[1103](https://app.zenhub.com/workspaces/team-volt-60a61059028fd7000e49956e/issues/gh/energinet-datahub/team-volt/1103)


source/core/tests/migrations/test_views.py
```diff
@@ -227,3 +228,106 @@ def test__capacity_settlement_v1__when_metering_point_type_is_not_valid_ones__sh
 
     # Assert
     assert actual.count() == 2
+
+
+### Current vuew tests
```
ShaBloze, line 233:
```suggestion
### Current view tests
```

is this comment necessary?

x-platformcoder: not really

ShaBloze (approved):

PR #492, opened 2025-03-25
fix: Deploy both databricks when changes are made

ShaBloze:
# Description

When there was made changes in both python and dotnet, then only dotnet was deployed. We would like both to be deployed.


jpm-halfspace (approved):
This has been fixed in `python-uv-cd`

djorgensendk (approved):

PR #494, opened 2025-03-25
Measurements/Feat: Add Dependabot Yml

Gizmooa:
This pull request introduces a new `dependabot` configuration to automate dependency updates for various package ecosystems in the repository. The configuration specifies the update schedule, assignees, and commit message prefixes for each ecosystem.

Dependabot configuration:

* Added a `dependabot.yml` file to automate dependency updates for `uv` and `nuget` package ecosystems in different directories. The updates are scheduled weekly at 08:00 in the Europe/Copenhagen timezone.


.github/dependabot.yml
```diff
@@ -0,0 +1,49 @@
+version: 2
+updates:
+  - package-ecosystem: uv
+    directory: /source/core
+    schedule:
+      interval: weekly
```
BjarkeMeier, line 6: Considering the impact and our general security level: How about monthly?

.github/dependabot.yml
```diff
@@ -0,0 +1,49 @@
...
+      time: 08:00
+      timezone: Europe/Copenhagen
+    assignees:
+      - awha86
+      - Gizmooa
+      - HenrikSommer
+      - jonasdmoeller
+      - prtandrup
+      - ShaBloze
+      - x-platformcoder
```
BjarkeMeier, line 16: Does it support teams as-well? Could this be replaced by "team-volt"?

.github/dependabot.yml
BjarkeMeier, line 1: Does this include minors? I'd prefer not.

BjarkeMeier, line 1: Do security update checks also run weekly? Can we have a separate frequency for security and regular updates?

.github/dependabot.yml
Gizmooa: Sadly no, I was looking into it but it is not supported.

.github/dependabot.yml
Gizmooa: Reason why I set weekly, was because I thought it maybe could be used in "drift"-rotation. So a new person every week is responsible for updates. 😄 

But we can easily change it for e.g., Mandalorian 😄

.github/dependabot.yml
Gizmooa: Yes it does, what would you want included? I just thought any update would make sense since we are "just" checking once a week, and maybe month. As there could be some small security patches in packages.

Gizmooa: Otherwise we can adjust [update-types](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#update-types-groups)

Gizmooa: I'll have a look, currently I think it checks for both [update-types](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#update-types-groups)

.github/dependabot.yml
BjarkeMeier: Let's do monthly and minor for Mandalorian.

ShaBloze (approved):

BjarkeMeier (approved):

PR #498, opened 2025-03-25
refac: Adding `pytest-mock`

ShaBloze:
# Description

More clean way of mocking


Gizmooa (approved):

Gizmooa (changes requested):
Retracts approve until we've figured out of the mocks leaks out to other tests. It seems like a mocked object will be mocked for any other tests as well using this:
https://stackoverflow.com/questions/41188982/mock-scope-goes-beyond-current-test

Gizmooa (approved):

PR #501, opened 2025-03-25
Test : nsg6 changes on simple scenario

Troels36:
# Description

Test: added measurement where the supply to the grid is greater than the consumption from grid.


Troels36:
Due to merge conflicts

PR #503, opened 2025-03-25
refac: Refactored seeding

henrik-t-caroe:
# Description

- Instead of using the seeding class, it now has a method in each test `seed_table`, which is responsible for seeding the specific subsystem table.

- Added a `execute_statement` wrapper to the base fixture


source/geh_calculated_measurements/tests/subsystem_tests/net_consumption_group_6/seed_table.py
```diff
@@ -0,0 +1,76 @@
...
+        '{child_metering_point_id}',
+        'net_consumption',
+        '{parent_metering_point_id}',
+        '{datetime(2022, 12, 31, 23, 0, 0)}',
+        '{datetime(2023, 12, 31, 23, 0, 0)}'
+    )
+    """)
+
+    for statement in statements:
+        job_fixture.databricks_api_client.execute_statement(
```
JesperGomes, line 74: could it make sense to have a 'execute_statement' method on the fixture, so that we don't need to deal with the api_client, warehouse etc.

BjarkeMeier (dismissed):

source/geh_calculated_measurements/tests/subsystem_tests/electrical_heating/seed_table.py
```diff
@@ -0,0 +1,24 @@
...
+def seed_table(
+    job_fixture,
+) -> None:
+    gold_table_rows = [
+        GoldTableRow(
+            metering_point_id=170000060000000201,
+            observation_time=datetime(2024, 11, 30, 23, 0, 0),
+            quantity=random.uniform(0.1, 10.0),
+        )
+        for i in range(1)
```
JesperGomes, line 20: why not just create a single row (not a list)

PR #510, opened 2025-03-26
refac: Improve separation of domain and application in capacity settlement

BjarkeMeier:


source/geh_calculated_measurements/src/geh_calculated_measurements/capacity_settlement/domain/model/ten_largest_quantities.py
```diff
@@ -17,9 +17,9 @@ def __init__(self, df: DataFrame):
 
     schema = T.StructType(
         [
-            #
-            # ID of the orchestration that initiated the calculation job
-            T.StructField(ContractColumnNames.orchestration_instance_id, T.StringType(), not nullable),
+            # #
```
JesperGomes, line 20: delete?

source/geh_calculated_measurements/src/geh_calculated_measurements/capacity_settlement/application/calculation.py
```diff
@@ -26,26 +39,70 @@ def execute_application(spark: SparkSession, args: CapacitySettlementArgs) -> No
...
-    calculated_measurements_repository.write_calculated_measurements(calculated_measurements)
+    calculated_measurements_repository.write_calculated_measurements(calculated_measurements_hourly)
 
     # Create a repository to write the calculations and ten largest quantities
     capacity_settlement_repository = CapacitySettlementRepository(spark, args.catalog_name)
 
     # Write the calculations
-    calculations = calculation_output.calculations
-    capacity_settlement_repository.write_calculations(calculations)
+    capacity_settlement_repository.write_calculations(calculations.df)
```
JesperGomes, line 72: Why ".df"?

BjarkeMeier: There are different approaches regarding when using wrappers and data frames. I changed to this to avoid some problems regarding circular dependencies for now. But it's something we should get back to.
Let's talk if you want a more detailed explanation.

source/geh_calculated_measurements/src/geh_calculated_measurements/capacity_settlement/domain/model/ten_largest_quantities.py
```diff
@@ -17,9 +17,9 @@ def __init__(self, df: DataFrame):
...
+            # # ID of the orchestration that initiated the calculation job
+            # T.StructField(ContractColumnNames.orchestration_instance_id, T.StringType(), not nullable),
```
BjarkeMeier, line 22:
```suggestion
```

JesperGomes (approved):

PR #520, opened 2025-03-26
refac: Reorganize files in missing measurements log

JesperGomes:
# Description
Before:
image

After: 
image


source/geh_calculated_measurements/src/geh_calculated_measurements/missing_measurements_log/infrastructure/__init__.py
```diff
@@ -1,6 +1,6 @@
-from .electricity_market.repository import Repository as ElectricityMarketRepository
+from .repository import Repository as MeteringPointPeriodsRepository
```
energinet-ajw, line 1: If we only have a single repository per infrastructure, do we need to rename it?

JesperGomes: hmm - good point

energinet-ajw (dismissed):

energinet-ajw (approved):

PR #524, opened 2025-03-26
fix: Convert DH2 Unit Enum values to DH3

slb-halfspace:
# Description
Turns out DH2 has a different string-value for the enum indicating Tonnes. 
In DH2 it is simply "T" and in DH3 it is now "Tonne" or "TONNE".

This PR replaces all existing values of "T" with "TONNE", and ensures the conversion happens from bronze->silver.


ShaBloze (approved):
LGTM

PR #534, opened 2025-03-28
feat: Activate measurements scenario test nsg6

johevemi:
# Description


johevemi:
will be included in #535

PR #543, opened 2025-03-31
refac: Move streams out of repository

ShaBloze:
# Description

Moving write_stream out of repository, as it has nothing to do with the repository.

Added stream classes instead.

## References


source/core/src/core/gold/infrastructure/streams/gold_measurements_stream.py
```diff
@@ -0,0 +1,36 @@
...
+            self.data_lake_settings, self.gold_container_name, GoldTableNames.gold_measurements
+        )
+
+        write_stream = (
+            source_table.writeStream.format("delta")
+            .queryName(query_name)
+            .option("checkpointLocation", checkpoint_location)
+        )
+
+        if StreamingSettings().continuous_streaming_enabled is False:
```
slb-halfspace, line 33: Could use apply_streaming_settings instead?

source/core/src/core/silver/application/streams/migrated_transactions.py
```diff
@@ -9,18 +8,18 @@
...
     SilverMeasurementsRepository,
 )
+from core.silver.infrastructure.streams.silver_measurements_stream import SilverMeasurementsStream
 
 
 def stream_migrated_transactions_to_silver() -> None:
     spark = spark_session.initialize_spark()
     bronze_migrated_transactions_repository = MigratedTransactionsRepository(spark)
-    silver_repository = SilverMeasurementsRepository()
+    silver_repository = SilverMeasurementsStream()
```
x-platformcoder, line 17: Why assign it to a local variable? Down in the batch operation, you use the imported class directly.

ShaBloze: changed to use the imported class directly

PR #548, opened 2025-03-31
feat: Implement dependency injection

ShaBloze:
# Description

## References


ShaBloze:
Decided that we do not want to implement dependency injection

PR #560, opened 2025-04-01
Fix interfering tests

BjarkeMeier:


source/geh_calculated_measurements/tests/conftest.py
```diff
@@ -13,6 +13,7 @@
 from tests.testsession_configuration import TestSessionConfiguration
 
 
+# TODO BJM: Should we remove this?
```
jpm-halfspace, line 16: Would be kinda awesome

source/geh_calculated_measurements/tests/data_products/conftest.py
```diff
@@ -10,13 +8,12 @@
...
-from geh_calculated_measurements.database_migrations.migrations_runner import migrate
-from geh_calculated_measurements.database_migrations.settings.catalog_settings import CatalogSettings
-from tests import drop_calculated_measurements_databases, ensure_calculated_measurements_databases_exist
+from geh_calculated_measurements.database_migrations.migrations_runner import _migrate
+from tests import SPARK_CATALOG_NAME, ensure_calculated_measurements_databases_exist
 
 
-@pytest.fixture(scope="module")
-def migrations_executed(spark: SparkSession, dummy_logging) -> Generator[None, None, None]:
+@pytest.fixture(scope="session")
```
jpm-halfspace, line 15: Should this live in the global `conftest?`

source/geh_calculated_measurements/tests/electrical_heating/settings/conftest.py
```diff
@@ -1,9 +1,12 @@
 import os
 
+from tests import SPARK_CATALOG_NAME
 
+
+# TODO BJM: Should we remove this?
 def pytest_runtest_setup() -> None:
```
jpm-halfspace, line 7: Probably.... At least, if the decision is that patching should happen where it's used, then it doesn't make sense to have this

jpm-halfspace: Be aware that tests will probably break when removing this

source/geh_calculated_measurements/tests/net_consumption_group_6/job_tests/conftest.py
```diff
@@ -11,6 +11,7 @@
 from tests.net_consumption_group_6.job_tests import get_test_files_folder_path
 
 
+# TODO BJM: Rely on a fixture to set up these shared tables?
 @pytest.fixture(autouse=True)
 def gold_table_seeded(spark: SparkSession) -> None:
```
jpm-halfspace, line 16: Regardless of what we do here, `autouse=True` should be removed and the fixtures should be used explicitly

source/geh_calculated_measurements/tests/data_products/conftest.py
BjarkeMeier: Apparently, it's only required by data product tests, which I believe is a good thing as long as it can stay that way.

source/geh_calculated_measurements/tests/electrical_heating/settings/conftest.py
BjarkeMeier: That's why I only left a comment in this PR 😄

source/geh_calculated_measurements/tests/net_consumption_group_6/job_tests/conftest.py
BjarkeMeier: I agree 👍  I'll make that change when I refactor the fixture.

JesperGomes (approved):
LGTM

PR #562, opened 2025-04-01
Abstract base stream class example 

ShaBloze:
# Description

Do we want to do this?


ShaBloze:
Decided not to implement. Closing.

PR #566, opened 2025-04-02
refac: Use data product from python package

ShaBloze:
# Description

## References


source/core/src/core/contracts/README.md
```diff
@@ -2,11 +2,6 @@
 
 This directory contains contracts.
 
-## Data products
```
prtandrup, line 5: Instead of removing this, would it make sense to reference their new location?

prtandrup (approved):

PR #573, opened 2025-04-03
chore: improves error messaging when log query fails

jpm-halfspace:
# Description

### DIFF
--- a/source/geh_calculated_measurements/src/geh_calculated_measurements/testing/utilities/job_tester.py
+++ b/source/geh_calculated_measurements/src/geh_calculated_measurements/testing/utilities/job_tester.py
@@ -89,7 +89,7 @@ class JobTestFixture:
             "message": "An error occurred while querying the logs.",
             "details": None,
         }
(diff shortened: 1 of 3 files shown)


source/geh_calculated_measurements/src/geh_calculated_measurements/testing/utilities/job_tester.py
BjarkeMeier, line 1: Can you add screenshots to the description so I can verify this PR, since there are no tests?

BjarkeMeier: Yes, but what do I get instead? A pretty message or "NoneType has no property 'Foo'"?

jpm-halfspace: Doesn't matter. I'll just close it

BjarkeMeier (dismissed):

BjarkeMeier (approved):

PR #576, opened 2025-04-03
set spark log level to Error

sf-halfspace:
# Description

Sets the logging level for spark to "error"

## References

user story: https://app.zenhub.com/workspaces/mandalorian-5fd22a1a59e4740018f5ab5a/issues/gh/energinet-datahub/team-mandalorian/724


source/geh_calculated_measurements/tests/capacity_settlement/job_tests/conftest.py
```diff
@@ -13,6 +13,7 @@
 
 @pytest.fixture(scope="session")
 def gold_table_seeded(spark: SparkSession) -> None:
+    spark.sparkContext.setLogLevel("ERROR")
```
jpm-halfspace, line 16: You should only need to do this once in the `spark` fixture

sf-halfspace: What is the spark fixture? :)

source/geh_calculated_measurements/tests/capacity_settlement/job_tests/test_capacity_settlement_job.py
```diff
@@ -15,6 +15,8 @@
 def test_execute(
     spark: SparkSession, gold_table_seeded: Any, migrations_executed: None, dummy_logging, monkeypatch
 ) -> None:
+    spark.sparkContext.setLogLevel("ERROR")
```
BjarkeMeier, line 18: Why is this needed when it's already set in the spark fixture?

source/geh_calculated_measurements/tests/conftest.py
```diff
@@ -65,6 +65,7 @@ def spark() -> Generator[SparkSession, None, None]:
     """
     Create a Spark session with Delta Lake enabled.
     """
+    _spark.sparkContext.setLogLevel("ERROR")
```
BjarkeMeier, line 68: This fixture looks a bit special because despite being session-scoped it will be executed multiple times when using pytest-xdist. It's a nit thing, but I it would be nice to implement this in the `get_spark_test_session()` function. Probably as an optional argument with default value `ERROR`.
That way you'll actually fix all repos at once (when they eventually upgrade to using the latest geh_common).

source/geh_calculated_measurements/tests/net_consumption_group_6/job_tests/test_net_consumption_group_6_job.py
BjarkeMeier, line 1: Looks like a file that should be reverted?

PR #579, opened 2025-04-03
test: Add calculated to gold subsystem test

ShaBloze:
# Description

### DIFF
--- a/source/core/tests/subsystem_tests/features/transmission_of_calculated_measurements.feature
+++ b/source/core/tests/subsystem_tests/features/transmission_of_calculated_measurements.feature
@@ -1,0 +1,6 @@
+Feature: Transmission of calculated measurements
+
+Scenario: Processing calculated measurements
+  Given valid calculated measurements
+  When inserted into the calculated measurements table
+  Then the calculated measurements are avaiable in the Gold Layer
--- a/source/core/tests/subsystem_tests/fixtures/calculated_measurements_fixture.py
+++ b/source/core/tests/subsystem_tests/fixtures/calculated_measurements_fixture.py
@@ -1,0 +1,43 @@
+from geh_common.databricks.databricks_api_client import DatabricksApiClient
+
+from tests.subsystem_tests.builders.calculated_measurements_row_builder import CalculatedMeasurementsRow
+from tests.subsystem_tests.settings.databricks_settings import DatabricksSettings
+
+
+class CalculatedMeasurementsFixture:
+    def __init__(self) -> None:
+        self.databricks_settings = DatabricksSettings()  # type: ignore
+
+        self.databricks_api_client = DatabricksApiClient(
+            self.databricks_settings.token,
+            self.databricks_settings.workspace_url,
+        )
+
+    def insert_calculated_measurements(self, row: CalculatedMeasurementsRow) -> None:
+        """
+        Inserts the calculated measurements into the databricks table.
+        """
+        query = f"""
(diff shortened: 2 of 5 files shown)


source/core/tests/subsystem_tests/features/transmission_of_calculated_measurements.feature
```diff
@@ -0,0 +1,6 @@
+Feature: Transmission of calculated measurements
```
prtandrup, line 1: Should we perhaps consider using `streaming of` or `persisting calculated measurements` instead of `Transmission of`?

source/core/tests/subsystem_tests/features/transmission_of_calculated_measurements.feature
```diff
@@ -0,0 +1,6 @@
...
+
+Scenario: Processing calculated measurements
+  Given a new valid calculated measurements
+  When inserted into the calculated measurements table
+  Then the calculated measurements are available in the Gold Layer
```
prtandrup, line 6: Plural or not?
Plural might make best sense businesswise, even though the test uses one row.

```suggestion
Scenario: Processing calculated measurements
  Given new valid calculated measurements
  When inserted into the calculated measurements table
  Then the calculated measurements are available in the Gold Layer
```

source/core/tests/subsystem_tests/step_definitions/test_transmission_of_calculated_measurements.py
```diff
@@ -0,0 +1,30 @@
...
+
+@when("inserted into the calculated measurements table")
+def _(calculated_measurements_row: CalculatedMeasurementsRow) -> None:
+    calculated_measurements_fixture = CalculatedMeasurementsFixture()
+    calculated_measurements_fixture.insert_calculated_measurements(calculated_measurements_row)
+
+
+@then("the calculated measurements are avaiable in the Gold Layer")
+def _(calculated_measurements_row: CalculatedMeasurementsRow, gold_layer_fixture: GoldLayerFixture) -> None:
+    gold_layer_fixture.assert_measurement_persisted(calculated_measurements_row.orchestration_instance_id)
```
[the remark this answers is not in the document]
ShaBloze: The orchestration_instance_id is dynamic and changes for every test. It is created in the `@given`

source/core/tests/subsystem_tests/fixtures/calculated_measurements_fixture.py
```diff
@@ -0,0 +1,46 @@
...
+            self.databricks_settings.token,
+            self.databricks_settings.workspace_url,
+        )
+
+    def insert_calculated_measurements(self, row: CalculatedMeasurementsRow) -> None:
+        """
+        Inserts the calculated measurements into the databricks table.
+        """
+        database_name = "measurements_calculated_internal"
+        table_name = "calculated_measurements"
```
awha86, line 21: Should these be defined somewhere else, in `.env`perhaps?

ShaBloze: I wish we could, but then we would have to hardcode the values, like here, in our CD pipeline. So I think it's fine to have it here for now.

source/core/tests/subsystem_tests/features/transmission_of_calculated_measurements.feature
prtandrup, line 1: Nice edit, can we rename the file as well as the test file to streaming?

PR #595, opened 2025-04-07
Protobuf Python Files change enums to integer

ShaBloze:
# Description

Example of protobuf python files

And that we can get the enum Int values instead of the string.


x-platformcoder (approved):
LGTM

x-platformcoder (approved):
lgtm

PR #603, opened 2025-04-08
refac: Current measurements table

energinet-ajw:
This PR adds the class Current Measurements Table, which is inherited from the class Table.
A test of the added class is also added.

The following PR will replace the current Current Measurements repository, including tests.
When the next table is implemented, we will look at common elements and see if anything can be simplified. 

Comments to the Table will also be added.


source/geh_calculated_measurements/src/geh_calculated_measurements/common/infrastructure/current_measurements/current_meaurements_table.py
```diff
@@ -0,0 +1,17 @@
+import pyspark.sql.types as t
+
+from geh_calculated_measurements.common.infrastructure.current_measurements.database_definitions import (
+    MeasurementsGoldDatabaseDefinition,
+)
+from geh_calculated_measurements.common.infrastructure.current_measurements.table import Table
+
+
+class CurrentMeasurementsTable(Table):
+    def __init__(self, catelog_name: str) -> None:
```
prtandrup, line 10:
```suggestion
    def __init__(self, catalog_name: str) -> None:
```

source/geh_calculated_measurements/src/geh_calculated_measurements/common/infrastructure/__init__.py
```diff
@@ -15,5 +18,6 @@
     "CalculatedMeasurementsDatabaseDefinition",
     "CalculatedMeasurementsRepository",
     "CurrentMeasurementsRepository",
+    "CurrentMeasurementsTable",
```
prtandrup, line 21: I might be missing the bigger context 😄 ... is it intended that we call it "Current"? The gold table has both current and older data, the view is Current only

JesperGomes (approved):

PR #609, opened 2025-04-09
fix: Handle unknown protobuf message

ShaBloze:
# Description

We found out that when receiving an unknonwn protobuf message, the stream will fail when trying to extract the version.

Changing mode to "PERMISSIVE", which means that the default value will then be `None`.

The message will go to the `invalid_submitted_transactions` table.


x-platformcoder (approved):
lgtm

PR #610, opened 2025-04-09
fix: Latest run is current day 

johevemi:
# Description


source/geh_calculated_measurements/src/geh_calculated_measurements/net_consumption_group_6/domain/daily.py
```diff
@@ -91,7 +92,24 @@ def calculate_daily(
...
+    # If all rows were filtered out, return an empty dataframe with the required schema
+    if filtered_cenc.count() == 0:
+        schema = T.StructType(
+            [
+                T.StructField(ContractColumnNames.metering_point_id, T.StringType(), True),
+                T.StructField(ContractColumnNames.date, T.DateType(), True),
+                T.StructField(ContractColumnNames.quantity, T.DecimalType(18, 3), True),
+            ]
+        )
+        spark = initialize_spark()
```
BjarkeMeier, line 107: Will this be the same spark session as the one provided for tests from the fixture `spark`?

source/geh_calculated_measurements/src/geh_calculated_measurements/net_consumption_group_6/domain/daily.py
```diff
@@ -91,7 +92,24 @@ def calculate_daily(
         )
     )
 
-    df = cenc_w_last_run.select(
+    # Filter out rows where last_run is >= execution_start_datetime
+    filtered_cenc = cenc_w_last_run.filter(F.col("last_run") < F.col("execution_start_datetime"))
```
BjarkeMeier, line 96: What is the "w"?

johevemi: "with" I guess

source/geh_calculated_measurements/src/geh_calculated_measurements/net_consumption_group_6/domain/daily.py
```diff
@@ -91,7 +92,24 @@ def calculate_daily(
...
+
+    # If all rows were filtered out, return an empty dataframe with the required schema
```
BjarkeMeier, line 98: How about returning `None`?

source/geh_calculated_measurements/src/geh_calculated_measurements/net_consumption_group_6/domain/daily.py
johevemi: Not sure, it is the same as used in entry_point

source/geh_calculated_measurements/src/geh_calculated_measurements/net_consumption_group_6/domain/daily.py
johevemi: that comes with a bunch of other problems, but we can do if that is the route we want to take

BjarkeMeier: And why do you return? The materialization may be expensive.

johevemi: ?? if I dont return, it does not work

source/geh_calculated_measurements/src/geh_calculated_measurements/net_consumption_group_6/domain/daily.py
johevemi: will go over naming in another pr

source/geh_calculated_measurements/src/geh_calculated_measurements/net_consumption_group_6/domain/daily.py
johevemi: after a talk it made sense to remove the whole if block

ClausQAEnerginet (approved):

PR #626, opened 2025-04-10
skipping scenario test under proper circumstances

sf-halfspace:
# Description

list of tests that are excluded when running locally:

- tests/subsystem_tests
- tests/integration

## References

user story: https://app.zenhub.com/workspaces/mandalorian-5fd22a1a59e4740018f5ab5a/issues/gh/energinet-datahub/team-mandalorian/734


source/geh_calculated_measurements/tests/conftest.py
```diff
@@ -38,6 +38,17 @@
 from tests.testsession_configuration import TestSessionConfiguration
 
 
+# https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_collection_modifyitems
+def pytest_collection_modifyitems(config, items) -> None:
```
BjarkeMeier, line 42: But the tests can and should be able to run locally!
So it should only be a question about whether they are executed _by default_.

sf-halfspace: These changes should ensure that the test only runs under the right circumstances

BjarkeMeier: According to your title the tests can not work locally.

source/geh_calculated_measurements/tests/conftest.py
```diff
@@ -36,9 +36,25 @@
...
     create_job_environment_variables,
 )
+from tests.subsystem_tests.environment_configuration import EnvironmentConfiguration
 from tests.testsession_configuration import TestSessionConfiguration
 
 
+# https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_collection_modifyitems
+def pytest_collection_modifyitems(config, items) -> None:
+    skip_subsystem_tests = pytest.mark.skip(
+        reason="Skipping subsystem tests because environmental variables could not be set"
```
johevemi, line 46:
```suggestion
        reason="Skipping subsystem tests because environmental variables could not be set.  See .sample.env file on how to set environmental variables locally"
```

johevemi (approved):

PR #632, opened 2025-04-10
Add test case for missing measurements log job

BjarkeMeier:
Introduce a test case for the missing measurements log job to ensure proper functionality and coverage. This includes removing the skip marker from the test.


BjarkeMeier:
Superseeded.

PR #643, opened 2025-04-12
test: Ensure that measurements core can stream from calculated measurements

BjarkeMeier:
Enhance test coverage for data availability in gold and improve naming conventions in seed table functions for better clarity. Update dependencies to ensure compatibility.


BjarkeMeier:
Task given to Henrik

PR #647, opened 2025-04-14
refac: Electrical heating now reads from delta and changed overwrite to append

henrik-t-caroe:
# Description

- Electrical heating now reads from delta instead of csv.
- Changed overwrite to append when seeding


source/geh_calculated_measurements/src/geh_calculated_measurements/electrical_heating/infrastructure/__init__.py
```diff
@@ -1,4 +1,6 @@
-from .repository import Repository as ElectricityMarketRepository
+from .repository import (
+    Repository as ElectricityMarketRepository,  # TODO JVM: Is the naming right? why not electrical heating?
```
BjarkeMeier, line 2: Because it's a repository for reading from EC.

PR #649, opened 2025-04-15
[MeasurementsApi] Add Unit to AggregatedMeasurementDto

HenrikSommer:
# Description

DataHub 3 UI needs the Unit for aggregated measurements

## References

https://github.com/energinet-datahub/team-volt/issues/1109


source/dotnet/Measurements.Application/Responses/GetAggregatedMeasurementsResponse.cs
```diff
@@ -9,6 +9,7 @@ namespace Energinet.DataHub.Measurements.Application.Responses;
 
 public class GetAggregatedMeasurementsResponse
 {
+    // ReSharper disable once AutoPropertyCanBeMadeGetOnly.Global - used by System.Text.Json
```
Gizmooa, line 12: How come this is needed now?

HenrikSommer: Because Resharper / Rider suggests to remove the init; but if we do, then (de)serialization doesn't work

Gizmooa (approved):
If no documentation for the response type should be updated, e.g., swagger, then lgtm :)

PR #652, opened 2025-04-15
[MeasurementsApi] Versioning

HenrikSommer:
# Description

Versioning of the MeasurementsApi

## References

https://github.com/Energinet-DataHub/team-volt/1138


source/dotnet/Measurements.WebApi.IntegrationTests/Controllers/MeasurementsControllerTests.cs
```diff
@@ -136,11 +136,48 @@ public async Task GetAggregatedMeasurementsAsync_WhenMeteringPointDoesNotExist_R
         Assert.Equal(HttpStatusCode.NotFound, actualResponse.StatusCode);
     }
 
+    [Fact]
```
Gizmooa, line 139: What if no version is given? I suppose it'll die, if so we will introduce a breaking as we were discussing? 😄 But there is not really a workaround so 😄

source/dotnet/Measurements.WebApi/Controllers/MeasurementsController.cs
```diff
@@ -7,14 +8,14 @@
...
 [Authorize]
-[Route("measurements")]
 public class MeasurementsController(IMeasurementsHandler measurementsHandler)
     : ControllerBase
 {
-    [HttpGet]
-    [Route("forPeriod")]
+    [MapToApiVersion(1)]
+    [HttpGet("forPeriod")]
     public async Task<IActionResult> GetMeasurementsAsync([FromQuery] GetMeasurementRequest request)
```
Gizmooa, line 19: I think we should have a suffix V1 on the method, to enable easy addition of a new method that supports V2

source/dotnet/Measurements.WebApi.IntegrationTests/Controllers/MeasurementsControllerTests.cs
HenrikSommer: "v1" is automatically prefixed using MapGroup in StartUp.cs

source/dotnet/Measurements.WebApi/Controllers/MeasurementsController.cs
HenrikSommer: Done by magic ;-)

Gizmooa (approved):

PR #662, opened 2025-04-22
Enhance job testing and update configurations

BjarkeMeier:
Improve pytest options, enhance orchestration instance ID handling in job tests, update dependencies, and reenable a previously skipped test for data writing to delta.


source/geh_calculated_measurements/src/geh_calculated_measurements/testing/utilities/job_tester.py
```diff
@@ -120,23 +134,38 @@ def _error_from_response(self, response: LogsQueryResult | LogsQueryPartialResul
...
+        """Execute a query and wait for its completion.
+
+        The query must return at least one row to be considered successful.
+        """
+        start_time = time.time()
+        elapsed_time = 0
+
         response = self.azure_logs_query_client.query_workspace(
-            self.azure_log_analytics_workspace_id, query, timespan=None
+            self.azure_log_analytics_workspace_id, query, timespan=timedelta(minutes=60)
```
jpm-halfspace, line 148: I found that this can be a little problematic as time zones are different between us and Azure, meaning that the 60 minutes we specify here might already have passed (since Azure is UTC and currently 2 hours behind)

source/geh_calculated_measurements/src/geh_calculated_measurements/testing/utilities/job_tester.py
```diff
@@ -120,23 +134,38 @@ def _error_from_response(self, response: LogsQueryResult | LogsQueryPartialResul
...
 
-            --- Error ---
-            {error}
-            """
-            raise ValueError(error_msg)
+        while response.status == LogsQueryStatus.SUCCESS and elapsed_time < timeout_minutes * 60:
+            if len(response.tables[0].rows) > 0:
+                return response
+            response = self.azure_logs_query_client.query_workspace(
+                self.azure_log_analytics_workspace_id, query, timespan=timedelta(minutes=60)
```
jpm-halfspace, line 155: This is unnecessary. The logs client waits for response, so we could implement retries... But just blasting the service with requests wont help anything.

source/geh_calculated_measurements/src/geh_calculated_measurements/testing/utilities/job_tester.py
```diff
@@ -162,35 +191,44 @@ def test__and_then_job_telemetry_is_created(self, fixture: JobTestFixture):
...
+        | where Properties["orchestration_instance_id"] == '{fixture.orchestration_instance_id}'
         """
 
         # Act
         actual = fixture.wait_for_log_query_completion(query)
 
         # Assert
         assert actual.status == LogsQueryStatus.SUCCESS, (
-            f"The query did not complete successfully: {actual.status}. Query: {query}"
+            f"Expected log entries, but none found: {actual.status}. Query: {query}"
```
jpm-halfspace, line 202: This is not necessarily the case. Non success could also mean many other things

source/geh_calculated_measurements/src/geh_calculated_measurements/testing/utilities/job_tester.py
```diff
@@ -50,6 +60,8 @@ def _get_job_by_name(self, job_name: str) -> BaseJob:
 
     def start_job(self) -> Wait[Run]:
         base_job = self._get_job_by_name(self.job_name)
+        if base_job.job_id is None:
```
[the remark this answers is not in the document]
BjarkeMeier: You might be right, but the tools don't know:
image

source/geh_calculated_measurements/src/geh_calculated_measurements/testing/utilities/job_tester.py
BjarkeMeier: That sounds like a bug in the client if that's true as I provide no timestamps but solely a delta.

source/geh_calculated_measurements/src/geh_calculated_measurements/testing/utilities/job_tester.py
BjarkeMeier: The problem is that I have experienced that logs are first available "long" (minutes) after the jobs have terminated. So each call succeeds immediately, but that doesn't help when the logs are not available.

But you're sure right about blasting the service. Thx. I'll add a sleep 👍

source/geh_calculated_measurements/src/geh_calculated_measurements/testing/utilities/job_tester.py
BjarkeMeier: In that case the status will not be SUCCESS. This is related to the refactoring of `wait_for_log_query_completion` because we need it to wait for the expected logs to be available - or timeout.

source/geh_calculated_measurements/src/geh_calculated_measurements/testing/utilities/job_tester.py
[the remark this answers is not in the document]
jpm-halfspace: Arhhh.... Everything makes sense :)

jpm-halfspace (approved):

PR #666, opened 2025-04-23
geh common package upgrade

sf-halfspace:
# Description

upgrading geh_common to version 5.8.5

Had to force protobuff to version 6 as streamline is not compatible with higher versions of protobuff.


source/core/pyproject.toml
```diff
@@ -32,7 +32,7 @@ dev = [
     "azure-eventhub>=5.14.0",
     "pytest-bdd>=8.1.0",
     "pytest-bdd-report>=1.0.1",
-    "protobuf>=6.30.1",
+    "protobuf>=3.20,<6",
```
jpm-halfspace, line 35: Did you check this with VOLT?

sf-halfspace: no but I will

jpm-halfspace (approved):

jpm-halfspace (approved):

PR #668, opened 2025-04-23
fix: Transaction Creation Date pruning filter for merges

slb-halfspace:
# Description
Every time we have a batch, it will have to consider every single row from that stream written to gold previously. This is a major time-sink, and by calculating which transaction_creation_dates a given batch touches, we can cut prune files efficiently to better merge data in our submitted / migrated streams to gold.


source/core/src/core/utility/delta_table_helper.py
```diff
@@ -54,13 +55,42 @@ def append_if_not_exists(
...
+    # For each interval, we create the filter for our clustering column.
+    filters = []
+    for interval in intervals:
+        if interval[0] == interval[1]:
+            filters.append(
+                f"CAST({current_alias_table_name}.{clustering_col} AS DATE) == '{interval[0].strftime('%Y-%m-%d')}'"
+            )
+        else:
+            filters.append(
+                f"(CAST({current_alias_table_name}.{clustering_col} AS DATE) >= '{interval[0].strftime('%Y-%m-%d')}' AND CAST({current_alias_table_name}.{clustering_col} AS DATE) <= '{interval[1].strftime('%Y-%m-%d')}')"
```
prtandrup, line 91: Sure it should be < and = to, and not just <?
```suggestion
                f"(CAST({current_alias_table_name}.{clustering_col} AS DATE) >= '{interval[0].strftime('%Y-%m-%d')}' AND CAST({current_alias_table_name}.{clustering_col} AS DATE) < '{interval[1].strftime('%Y-%m-%d')}')"
```

slb-halfspace: Each interval made by the code above is inclusive, so the filters should be inclusive too. For a single date for instance, it would be a tuple of ("2025-01-01", "2025-01-01") which < would remove which is not intended.

ShaBloze (approved):

PR #678, opened 2025-04-25
test: subsystem for migrated silver measurements to measurements gold

x-platformcoder:
# Description

Create a subsystem test which inserts data into the Migration Silver Table and asserts that it lands in the Measurements gold table.

## References

[1134](https://app.zenhub.com/workspaces/team-volt-60a61059028fd7000e49956e/issues/gh/energinet-datahub/team-volt/1134)


awha86 (approved):
LGTM 👍

source/core/tests/subsystem_tests/features/streaming_of_migrated_measurements.feature
```diff
@@ -0,0 +1,6 @@
+Feature: Stream migrated transaction into Measurements gold
+
+Scenario: Processing migrated transaction
+  Given a new valid migrated transaction
+  When inserted into the Migration silver table
+  Then the migrated transaction is available in the Gold layer
```
ShaBloze, line 6:
```suggestion
  Given a new valid migrated transaction inserted into the Migration Silver Table
  When streaming from Nigration silver to Measurements gold
  Then the migrated transaction is available in the Gold layer
```

Because our "when" is different compared to the other scenarios we have. I think it is OK that our "given" appends to the Migration silver table.

PR #693, opened 2025-04-28
[Measurements API] Enum parser cleanup

jonasdmoeller:
# Description


source/dotnet/Measurements.UnitTests/Application/Responses/EnumParserTests/UnitParserTests.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    [Theory]
+    [InlineData("kwh", Unit.kWh)]
+    [InlineData("kw", Unit.kW)]
+    [InlineData("mw", Unit.MW)]
+    [InlineData("mwh", Unit.MWh)]
+    [InlineData("tonne", Unit.Tonne)]
+    [InlineData("kvarh", Unit.kVArh)]
+    [InlineData("mvar", Unit.MVAr)]
+    public void ParseUnit_ValidInput_ReturnsExpectedUnit(string input, Unit expected)
```
ShaBloze, line 19: should we add tests where uppercase appears in the input? like `kWh`. The api will receive these values : [quantity unit](https://github.com/Energinet-DataHub/opengeh-python-packages/blob/main/source/geh_common/src/geh_common/domain/types/quantity_unit.py)

jonasdmoeller: Good point, I added these test cases

ShaBloze (approved):
LGTM

PR #701, opened 2025-04-29
feat: Limit calculated measurements view to past 3 years

johevemi:
# Description


source/geh_calculated_measurements/src/geh_calculated_measurements/database_migrations/migration_scripts/20250429124000_recreate_view__calculated_measurements_v1.sql
```diff
@@ -0,0 +1,19 @@
+DROP VIEW IF EXISTS {catalog_name}.{calculated_measurements_database}.calculated_measurements_v1
+GO
+
+CREATE VIEW {catalog_name}.{calculated_measurements_database}.calculated_measurements_v1 AS
```
JesperGomes, line 4: CREATE OR REPLACE?

johevemi: i can never remember if it works or not

JesperGomes: ...or simply rename the existing script with current time stamp?

johevemi: the old script dropped a different table hourly_calculated_measurements_v1

JesperGomes: ok - but CREATE OR REPLACE should work, and AFAIK the recommended way to do it

johevemi: it is now using create or replace and the data product test passed locally

source/geh_calculated_measurements/tests/data_products/given_calculated_measurements/when/measurements_calculated_internal.calculated_measurements.csv
```diff
@@ -1,4 +1,27 @@
 orchestration_type;orchestration_instance_id;transaction_id;transaction_creation_datetime;metering_point_id;metering_point_type;observation_time;quantity;#comment
+electrical_heating;00000000-0000-0000-0000-000000000001;test_id;2025-03-29 23:00:00Z;140000000000170201;electrical_heating;2022-03-29 23:00:00Z;2.1;observation_time older than 3 year from current date
+electrical_heating;00000000-0000-0000-0000-000000000001;test_id;2025-03-29 23:00:00Z;140000000000170201;electrical_heating;2022-03-30 00:00:00Z;0;observation_time older than 3 year from current date
```
BjarkeMeier, line 3: Observation time cannot be newer than transaction creation time.

source/geh_calculated_measurements/tests/data_products/given_calculated_measurements/when/measurements_calculated_internal.calculated_measurements.csv
```diff
@@ -1,4 +1,27 @@
 orchestration_type;orchestration_instance_id;transaction_id;transaction_creation_datetime;metering_point_id;metering_point_type;observation_time;quantity;#comment
```
BjarkeMeier, line 1: This test doesn't verify that the cut off is at exactly 3 years.
In addition are there any details on whether it should be at the same time as "now" 3 years ago? Or should it be at midnight 3 years ago, or...?

source/geh_calculated_measurements/tests/data_products/given_calculated_measurements/when/measurements_calculated_internal.calculated_measurements.csv
johevemi: You mean transaction_creation_datetime cannot be later than observation_time. I will set the year to 2022

source/geh_calculated_measurements/tests/data_products/given_calculated_measurements/when/measurements_calculated_internal.calculated_measurements.csv
johevemi: Correct, it does not test the exact 3 year cut off date. It has been tested manually because it was a hassle to mock current_timestamp(), if you know how it can be done easily please let me know.

I don't know, it takes the entire day.

BjarkeMeier: I'm not saying that we must test it further. But here is an idea on how we _could_ do it:
Create unit tests that adds data to internal table and assert that they are either selected by the view or not. You can seed with dynamic data where observation time is selected around 3 years before "now". Like 2 min before and after just to stay clear of race conditions and latency in the tests. It should be fairly simple to do. And I would probably consider rolling back the scenario test changes.

source/geh_calculated_measurements/tests/data_products/given_calculated_measurements/when/measurements_calculated_internal.calculated_measurements.csv
BjarkeMeier: See my other comment as well.

source/geh_calculated_measurements/tests/data_products/given_calculated_measurements/when/measurements_calculated_internal.calculated_measurements.csv
johevemi: yeah okay that would test the 3 year mark, but we would still have a problem with the scenario test in about 3 years since current_timestamp() is not mocked

BjarkeMeier: Hmm, I see. Ok, so we'll have to update our scenario tests every 3 year.
Perhaps we should have a dev brainstorm and see what ideas we can come up with.

BjarkeMeier (dismissed):

source/geh_calculated_measurements/tests/data_products/conftest.py
```diff
@@ -10,11 +12,64 @@
...
+
+    return test_timestamp_str
+
+
 @pytest.fixture(scope="module")
 def test_cases(
     spark: SparkSession,
     request: pytest.FixtureRequest,
     migrations_executed: None,  # Used implicitly
+    execution_datetime,  # Use this instead of directly calculating it
```
BjarkeMeier, line 71:
```suggestion
```

source/geh_calculated_measurements/tests/data_products/conftest.py
```diff
@@ -10,11 +12,64 @@
 from tests import SPARK_CATALOG_NAME
 
 
+@pytest.fixture(scope="module")
+def execution_datetime(request: pytest.FixtureRequest):
```
BjarkeMeier, line 16: Turn this fixture into a plain function to simplify matters and save some memory?

johevemi: or maybe I should just get the execution_start_datetime directly in the override_current_timestamp fixture, because it is not used anywhere else, which I though it would be at first

BjarkeMeier: I considered that. But it's good that you restore the spark config, and I find this a nice way to do it. Otherwise, I think your override_current_timestamp will become more complex than desired.

johevemi: yeah okay, that makes sense. I will just change execution_datetime in to a function

source/geh_calculated_measurements/tests/data_products/conftest.py
```diff
@@ -10,11 +12,62 @@
...
+        execution_datetime: Datetime to use for current_timestamp()
+    """
+    # Ensure the timestamp has UTC timezone
+    test_timestamp = execution_datetime(request)
+    if test_timestamp.tzinfo is None:
+        test_timestamp = test_timestamp.replace(tzinfo=UTC)
+    test_timestamp_str = test_timestamp.strftime("%Y-%m-%d %H:%M:%S")
+
+    # Set timezone to UTC
+    spark.conf.set("spark.sql.session.timeZone", "UTC")
```
BjarkeMeier, line 52: Why did you remove the restoring of spark conf? Doesn't this possibly break other tests?

johevemi: I added it again. I thought it rest conf because of scope, but idk

BjarkeMeier (approved):

PR #726, opened 2025-05-01
feat; Add bronze integrationtest scenarios

ShaBloze:
# Description


source/core/tests/integration/features/stream_migration_to_bronze_migrated.feature
```diff
@@ -0,0 +1,11 @@
+Feature: Stream Migration from Silver to Bronze Measurements
+
+  Scenario: Daily Load: Migrate Transactions from Silver to Bronze Measurements
+    Given transactions available in the migration silver table
+    When streaming daily load from Migration silver to Measurements Bronze
+    Then transactions should be available in the bronze measurements migration table
```
ShaBloze, line 6: This is nice. 

Maybe we should consider making a new or change this a bit to:

`Given newer transactions available in the Migration silver table and the older has already been taken`

I think I would like to see that we have 2 rows in the Migration Silver Table and we only ingest the newer transaction, as we have already ingested the older one. Does it make sense?

example: 
- Create 1 row in Migration Silver Table which has a `created` date from yesterday.
- Create 1 row in Migration Silver Table which has a `created` date from 1 week ago
- The `When` should then set the `latest_created_already_migrated` to 2 days ago
- `Then` only expects a the newest row

source/core/tests/integration/step_definitions/test_stream_migration_to_bronze_migrated.py
```diff
@@ -66,6 +101,15 @@ def _(spark: SparkSession, mock_checkpoint_path):
 # Then steps
 
 
+@then("only the newer transaction should be available in the bronze measurements migration table")
+def _(spark: SparkSession, expected_newer_metering_point_id):
+    df = spark.table(f"{BronzeSettings().bronze_database_name}.{BronzeTableNames.bronze_migrated_transactions_table}")
+    count = df.filter(f"metering_point_id = '{expected_newer_metering_point_id}'").count()
+    total = df.count()
+    assert count == 1, f"Expected newer transaction '{expected_newer_metering_point_id}' to be migrated"
+    assert total == 1, f"Expected only one transaction in the bronze table, but found {total}"
```
ShaBloze, line 110: Should we ensure that the "old" transaction is not in the table?

awha86 (approved):

PR #737, opened 2025-05-06
[MeasurementsApi] MeasurementsClient interface for process manager (PM28)

HenrikSommer:
# Description

Adds unimplemented methods for getting current and aggregated measurements for a specified period

## References

[PM28](https://miro.com/app/board/uXjVLV8HbBg=/)


source/dotnet/Measurements.Abstractions/Api/Models/MeasurementPointDto.cs
```diff
@@ -12,7 +12,7 @@
 /// <param name="RegistrationTime">Timestamp defining when this point was registered.</param>
 public sealed record MeasurementPointDto(
     int Order,
-    decimal Quantity,
+    decimal? Quantity,
```
Gizmooa, line 15: What is the difference between giving the quantity, and not? Maybe this should be documented as well. And describe in the readme why it has been adjusted.

source/dotnet/Measurements.Client/MeasurementsClient.cs
```diff
@@ -30,6 +31,11 @@ public async Task<MeasurementDto> GetByDayAsync(GetByDayQuery query, Cancellatio
         return result ?? throw new InvalidOperationException("The response was not successfully parsed.");
     }
 
+    public Task<ReadOnlyCollection<MeasurementPointDto>> GetByPeriodAsync(GetByPeriodQuery query, CancellationToken cancellationToken = default)
+    {
+        throw new NotImplementedException();
```
Gizmooa, line 36: Shouldn't this be called inside our application, such that we can test the connectivity? So we still call the endpoint, but we would get a not implemented thrown back? 😄

PR #757, opened 2025-05-14
Measurement API: Fix bug 1266: missing estimated marker

jonasdmoeller:
# Description


source/dotnet/Measurements.Application/Responses/MeasurementsAggregatedByDateResponse.cs
```diff
@@ -55,10 +54,18 @@ private static Unit SetUnit(AggregatedMeasurementsResult aggregatedMeasurementsR
...
+        var resolutionValue = aggregatedMeasurements.Resolutions.Single().ToString()
+                              ?? throw new InvalidOperationException();
+        var resolution = ResolutionParser.ParseResolution(resolutionValue);
+
+        var expectedPointCount = resolution switch
+        {
+            Resolution.QuarterHourly => 96,
+            Resolution.Hourly => 24,
+            Resolution.Daily => 1,
+            Resolution.Monthly or Resolution.Yearly => 0,
```
HenrikSommer, line 66: When normal time is changed to DST, only 92 points are expected for quarterhourly measured metering points, and vice versa when DST is changed to normal time, 100 points are expected. The expected numbers for hourly measured metering points are 23 and 25.
This is why the expected points was calculated based on the number of hours between the first and last observation of the day times 4 if quarterhourly measured, or just 1 when hourly measured (`ResolutionExtensions.GetExpectedPointCount`)

source/dotnet/Measurements.Application/Extensions/ResolutionExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        var duration = zonedEnd.ToInstant() - zonedStart.ToInstant();
+
+        return resolution switch
+        {
+            Resolution.Hourly => (int)duration.TotalHours,
+            Resolution.QuarterHourly => (int)(duration.TotalMinutes / 15),
+            Resolution.Daily => (int)duration.TotalDays,
+            Resolution.Monthly => 0,
+            Resolution.Yearly => 0,
```
HenrikSommer, line 26: I'm unsure about these - should they be 0 or 1?

jonasdmoeller: Lets go with your initial guess of 1 and fix eventually

HenrikSommer (approved):

PR #768, opened 2025-05-19
feat: Add `optimize_table` entry point

ShaBloze:
# Description

Refactoring:
- Adding a `databases` folder which contains `migrations` and `optimization`

## References

- https://github.com/Energinet-DataHub/team-volt/issues/1122


source/core/src/core/databases/optimization/optimize_table.py
```diff
@@ -0,0 +1,12 @@
...
+
+def optimize_table(database: str, table: str) -> None:
+    """Optimize a table in the database.
+
+    Args:
+        database (str): The name of the database.
+        table (str): The name of the table to optimize.
+    """
+    spark = spark_session.initialize_spark()
+    spark.sql(f"OPTIMIZE {database}.{table}")
```
Gizmooa, line 12: I'd also Display it, such that we can see how many files have been hit etc.

Gizmooa (approved):

PR #774, opened 2025-05-20
Measurements/Core/Feat: Add publish of swagger

Gizmooa:
This pull request introduces functionality to generate and publish a Swagger file for the Measurements API.. The key updates include adding a post-build target for Swagger generation, and configuring a CLI tool for Swagger.

Some guards had to be disabled, as PostConfiguration would not be possible otherwise, we might find a workaround later. But anyhows, the system will fail during an IntegrationTest or AcceptanceTest if the guards fails.


source/dotnet/Measurements.WebApi/.config/dotnet-tools.json
```diff
@@ -0,0 +1,12 @@
+{
+    "version": 1,
```
HenrikSommer, line 2:
```suggestion
    "version": 3,
```

HenrikSommer (approved):

HenrikSommer (approved):

PR #813, opened 2025-06-06
Alter `current_sap_v1` view to only show orchestration type `migration`

ShaBloze:
# Description

Forgot to only return measurements with orchestration type `migration`.

## References

- https://github.com/Energinet-DataHub/team-volt/issues/1194


ShaBloze:
Closing, we will read all for now.

PR #814, opened 2025-06-06
feat: Add migrated to SAP series stream

ShaBloze:
# Description

## References

- https://github.com/Energinet-DataHub/team-volt/issues/1208


docs/core/images/measurements_core_architecture.drawio.svg
ShaBloze, line 1: It might be confusing when we will add the calculated -> SAP.

Should we instead make 3 drawings, one for each "source"?

x-platformcoder: @prtandrup this is your ballgame, what do you suggest

source/core/src/core/gold/application/streams/migrated_sap_series_stream.py
```diff
@@ -0,0 +1,32 @@
+from pyspark.sql import DataFrame
+
+import core.bronze.application.config.spark_session as spark_session
+import core.gold.domain.transformations.sap_series_transformations as series_sap_transformations
```
x-platformcoder, line 4: sap_series_transformation :-D

x-platformcoder (dismissed):

x-platformcoder (approved):

PR #841, opened 2025-06-12
feat: Add cluster keys and adjust SAP series append

ShaBloze:
# Description

## References

- https://github.com/Energinet-DataHub/team-volt/issues/1215


source/core/tests/helpers/builders/gold_sap_series_builder.py
```diff
@@ -0,0 +1,43 @@
+from datetime import datetime
```
x-platformcoder, line 1: there is already a helper for this SAPSeriesBuilder

ShaBloze: oh I missed that. Will remove this one

PR #847, opened 2025-06-16
feat: Add datetime cluster key for sap series merge

ShaBloze:
# Description

## References

- https://github.com/Energinet-DataHub/team-volt/issues/1215


Gizmooa (approved):
Could consider a verbose flag, simply to avoid having all streams to print when they use the delta_table_helper. But this is not anything related to the functionality itself, so you can decide if you see it fit. Lgtm

PR #853, opened 2025-06-17
fix: orchestration_type added to sap view contracts

x-platformcoder:
# Description

Missing orchestaration_type


ShaBloze (approved):
LGTM

